PR #71, opened 2014-11-18
Added a GetEnterpriseUserAsync method to BoxUsersManager

saguiitay:
This will allow administrators to retrieve information about a specific user


cburnette:
just added As-User support:  https://github.com/box/box-windows-sdk-v2/blob/master/README.md#make-api-calls-with-as-user

PR #97, opened 2015-04-10
Fixed create shared link to take default null value

brianytang:


boxcla:
Verified that @brianytang has signed the CLA. Thanks for the pull request!

Issue #106, opened 2015-08-04
Implementing Box Developer Edition


bgoldman:
CODE CHANGES

Existing class BoxClient:

1. Create new public static method GetAppAuthConnection: Args for enterpriseId, config (BoxConfig), privateKey, privateKeyPassword. Requests enterprise token using new AuthRepository.AuthenticateAsync method. Instantiates and returns new BoxClient.
2. Create new public static method GetAppUserConnection: Args for userId, config (BoxConfig), privateKey, privateKeyPassword. Requests user token using new AuthRepository.AuthenticateAsync method. Instantiates and returns new BoxClient.

Existing class AuthRepository:

1. Create new public AuthenticateAsync method. Args for entityId, type, private key. Creates a JWT assertion and then uses that assertion to request an auth token. Called by BoxClient.GetAppAuthConnection and BoxClient.GetAppUserConnection.
2. Create new private method ReadPrivateKey: No args. Converts the private key string and returns SigningCredentials instance, decrypting if necessary. Called by AuthenticateAsync.

Existing class BoxUsersManager:

1. Create new public method CreateEnterpriseUserAsync: Args for userRequest (BoxUserRequest), fields (List<string>). Creates a new user.
2. Create new public method CreateAppUser: Same args as CreateEnterpriseUserAsync. Calls CreateEnterpriseUserAsync, setting is_platform_access_only=true.

NOTES

1. Creating an app user will be done by pass-through using BoxUsersManager.CreateEnterpriseUserAsync, without needing to modify this method beyond adding one param. It will authenticate using a BoxClient instance created from GetAppAuthConnection. Since BoxUsersManager.CreateEnterpriseUserAsync doesn’t exist, that will be created as well.

2. Getting an app user will be done using client.UsersManager, without needing to modify any methods. It will authenticate using a BoxClient instance created from GetAppUserConnection.

3. Getting an app user’s info will be done using client.UsersManager, without needing to modify any methods. It will authenticate using a BoxClient instance created from GetAppUserConnection.

4. Editing an app user will be done using client.UsersManager.UpdateUserInformationAsync, without needing to modify this method. It will authenticate using a BoxClient instance created from GetAppUserConnection.

5. Deleting an app user will be done using client.UsersManager.DeleteEnterpriseUserAsync, without needing to modify this method. It will authenticate using a BoxClient instance created from GetAppUserConnection.

Rod-Sychev:
Brandon, 
Is this for creating users on the fly?

[...]

PR #109, opened 2015-09-13
Box Developer Edition, part one

bgoldman:
Implementing https://github.com/box/box-windows-sdk-v2/issues/106

This is incomplete. I'm having serious issues getting the packages installed for the .NET version target (4.0) and needed to change it to 4.5. I kind of got it to work, but then a bunch of other things broke that shouldn't have broken.

The test project worked perfectly fine, but the integration is having these problems.

### DIFF
--- a/Box.V2.Test.Integration/BoxResourceManagerTestIntegration.cs
+++ b/Box.V2.Test.Integration/BoxResourceManagerTestIntegration.cs
@@ -15,6 +15,10 @@ namespace Box.V2.Test.Integration
         public const string ClientId = "YOUR_CLIENT_ID";
         public const string ClientSecret = "YOUR_CLIENT_SECRET";
 
+        public const string EnterpriseId = "YOUR_ENTERPRISE_ID";
+        public const string PrivateKey = "YOUR_PRIVATE_KEY";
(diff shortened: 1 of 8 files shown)


boxcla:
Verified that @bgoldman has signed the CLA. Thanks for the pull request!

varungu:
@bgoldman I will try to fix it today. Can you explain what all this change do and what is pending?

varungu:
@bgoldman The problem is that you are using "BouncyCastle", which is not compatible with portable-net4. You should instead use "Portable.BouncyCastle". Please try it out and let me know if it fixes your issues.

We cannot change the target to .net4.5, this will break this SDK for lot of platforms.

cburnette:
hi @varungu @bgoldman , you are correct that we can't change the target platform for this SDK.  The changes by @bgoldman are to support Box Platform Edition, but I have a separate branch going to roll these changes (and a few others) in and keep the targets the same.

bgoldman:
@varungu I have already tried Portable.BouncyCastle prior to submitting, but had issues with the other packages. I changed it back to the regular BouncyCastle after I changed the .NET version target.

@cburnette If you're able to keep the targets the same but still these packages in there and the code working, that would be great :)

cburnette:
Box Platform support has been added via a different branch.

PR #113, opened 2015-10-14
Jwt auth

cburnette:
changes to Box.V2 to support JWT and App users.  Also includes two new projects: Box.V2.JWTAuth and Box.V2.Samples,JWTAuth


boxcla:

PR #119, opened 2015-11-05
Download url

cburnette:
added method to get the direct Uri of a file;
also added support for As-User


boxcla:

PR #125, opened 2015-12-18
Retention Policy bug fixes

cburnette:
added missing 'Async' to some methods;
made RetentionLength property nullable for request;
made RetentionLength a string in response class because Box will return either an int or a string;


boxcla:

PR #133, opened 2016-03-09
Added a default user agent string to BoxConfig

brianytang:


boxcla:
Verified that @brianytang has signed the CLA. Thanks for the pull request!

varungu:
+1

PR #154, opened 2016-08-16
Pre flight upload url

cburnette:
Adds support for accessing the data returned from pre-flight check


boxcla:

PR #165, opened 2016-08-31
Resource: Files, FileManager

marek-vysoky:
Changes and comments, unit tests for:
Delete File
Download File
Get Embed Link
Get File's Info
Get File's Tasks
Get Thumbnail
Copy File
Create Shared Link
Download Version
Get Trashed File
Lock and Unlock
Permanently Delete
Preflight Check
Restore Item
Update File
Update File Info
Upload File
View Comments

### DIFF
--- a/Box.V2.Test.Integration/BoxFilesManagerTestIntegration.cs
+++ b/Box.V2.Test.Integration/BoxFilesManagerTestIntegration.cs
@@ -167,7 +167,7 @@ namespace Box.V2.Test.Integration
                 Name = copyName,
                 Parent = new BoxRequestEntity() { Id = "0" }
             };
-            BoxFile fileCopy = await _client.FilesManager.CopyAsync(copyReq);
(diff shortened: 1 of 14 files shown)


boxcla:

cburnette:
Hi Marek, a couple things here.  This pull request is too large and too wide-ranging.  Can you please submit a pull request, for example, that just contains comment changes?  Also, we cannot do things like change the method signatures of public methods (https://github.com/box/box-windows-sdk-v2/pull/165/files#diff-31d54c9ff3d390e24ad59beda04dd6acL268).  You've also removed some functionality that I'm not sure you understand it's purpose (https://github.com/box/box-windows-sdk-v2/pull/165/files#diff-31d54c9ff3d390e24ad59beda04dd6acL157).  This SDK is used by a large number of people and we have to be very precise when changing anything.  We cannot break backwards compatibility at this point in time.

PR #167, opened 2016-09-02
fix for building Uri's when the existing Uri contains a query string

cburnette:


boxcla:

PR #191, opened 2016-09-06
Feature/restore folder

valer-kovac:
RestoreTrashedFolderAsync - Updated validation of input parameters

### DIFF
--- a/Box.V2.Test/BoxFilesManagerTest.cs
+++ b/Box.V2.Test/BoxFilesManagerTest.cs
@@ -7,6 +7,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using System.Net.Http.Headers;
+using System.IO;
 
 namespace Box.V2.Test
 {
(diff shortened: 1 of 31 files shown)


boxcla:
Verified that @valer-kovac has signed the CLA. Thanks for the pull request!

Box.V2.Test/BoxFilesManagerTest.cs
```diff
@@ -189,6 +189,7 @@ public BoxFilesManagerTest()
 
             BoxFileRequest request = new BoxFileRequest()
             {
+                Id="5000948880",
```
coolboy: space between =

Box.V2/Managers/BoxFoldersManager.cs
```diff
@@ -264,19 +264,32 @@ public BoxFoldersManager(IBoxConfig config, IBoxService service, IBoxConverter c
...
                 .Id.ThrowIfNullOrWhiteSpace("folderRequest.Parent.Id");
 
             BoxRequest request = new BoxRequest(_config.FoldersEndpointUri, folderRequest.Id)
                     .Method(RequestMethod.Post)
-                    .Param(ParamFields, fields)
-                    .Payload(_converter.Serialize(folderRequest));
+                    .Param(ParamFields, fields);
+
+            // ID shall not be used in request body it is used only as url attribute
+            string oldId = folderRequest.Id;
```
coolboy: I am not fully understand why you are going the resetting here, could you explain?

valer-kovac: I decided to rememeber and renew ID value after processing to let it be transparent from caller point of view. I think it is not very good to change value of outside data internally may influence caller in some cases if he decides to use this value somehow again after method call.

coolboy:
@marek-vysoky conflict

marek-vysoky:
Conflict resolved

cburnette:
@coolboy Needs review

PR #210, opened 2016-09-22
Advanced Search, Search for query

marek-vysoky:
Adding comments

### DIFF
--- a/Box.V2/Managers/BoxSearchManager.cs
+++ b/Box.V2/Managers/BoxSearchManager.cs
@@ -23,12 +23,12 @@ namespace Box.V2.Managers
         /// <summary>
         /// Search for items that are accessible by a single user or an entire enterprise.
         /// </summary>
-        /// <param name="keyword">The string to search for; can be matched against item names, descriptions, text content of a file, and other fields of the different item types.</param>
-        /// <param name="limit">Number of search results to return</param>
-        /// <param name="offset">The search result at which to start the response</param>
-        /// <param name="fields"></param>
+        /// <param name="keyword">The string to search for. Box matches the search string against object names, descriptions, text contents of files, and other data.</param>
+        /// <param name="limit">Number of search results to return. The default is 30 and the maximum is 200.</param>
+        /// <param name="offset">The search result at which to start the response. The default is 0.</param>
+        /// <param name="fields">Attribute(s) to include in the response.</param>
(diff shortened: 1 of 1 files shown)


boxcla:

Box.V2/Managers/BoxSearchManager.cs
```diff
@@ -63,6 +63,7 @@ public BoxSearchManager(IBoxConfig config, IBoxService service, IBoxConverter co
                                                              
         {
 
+            keyword.ThrowIfNullOrWhiteSpace("keyword");
```
marek-vysoky: https://docs.box.com/reference#searching-for-content  query is REQUIRED.
Can I change value string keyword = null, to string keyword,.?
Do you agree?

coolboy: I think it should be safe to remove = null, since no one should be using it w/o keyword.

My another question, why this param named keyword instead query to align with our API doc?

Box.V2/Managers/BoxSearchManager.cs
```diff
@@ -23,12 +23,12 @@ public BoxSearchManager(IBoxConfig config, IBoxService service, IBoxConverter co
...
-        /// <param name="fields"></param>
+        /// <param name="keyword">The string to search for. Box matches the search string against object names, descriptions, text contents of files, and other data.</param>
+        /// <param name="limit">Number of search results to return. The default is 30 and the maximum is 200.</param>
+        /// <param name="offset">The search result at which to start the response. The default is 0.</param>
+        /// <param name="fields">Attribute(s) to include in the response.</param>
         /// <param name="scope">The scope for which you want to limit your search to. Can be user_content for a search limited to only the current user or enterprise_content for the entire enterprise. To enable the enterprise_content scope for an administrator, please contact Box.</param>
-        /// <param name="fileExtensions">Limit searches to specific file extensions like pdf,png,doc</param>
+        /// <param name="fileExtensions">Limit searches to specific file extensions like pdf,png, or doc. The value can be a single file extension or a comma-delimited list of extensions. For example: png,md,pdf</param>
```
coolboy: ok

Box.V2/Managers/BoxSearchManager.cs
cburnette: @coolboy @marek-vysoky No, keyword CAN be null when you are doing a metadata-based query.  DO NOT change it from defaulting to null.

marek-vysoky:
I have removed validation to null for keyword

PR #214, opened 2016-10-05
Feature/metadata file

marek-vysoky:
Metadata on File - Create, Get All, Get, Update metadata
Adding comments, creating unit tests

### DIFF
--- a/Box.V2/Extensions/BoxExtensions.cs
+++ b/Box.V2/Extensions/BoxExtensions.cs
@@ -34,5 +34,13 @@ namespace Box.V2.Extensions
 
             return value;
         }
+
+        internal static T? ThrowIfNullEnum<T>(this T? param, string name)  where T : struct
+        {
+            if (param == null)
+                throw new ArgumentNullException(name);
+
+            return param;
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


boxcla:

Box.V2/Extensions/BoxExtensions.cs
```diff
@@ -34,5 +34,13 @@ internal static string ThrowIfNullOrWhiteSpace(this string value, string name)
 
             return value;
         }
+
+        internal static T? ThrowIfNullEnum<T>(this T? param, string name)  where T : struct
```
coolboy: Why could not we remove the "where T : class" on ThrowIfNull(), so that we don't need ThrowIfNullEnum?

Box.V2/Managers/BoxMetadataManager.cs
```diff
@@ -170,6 +170,11 @@ public BoxMetadataManager(IBoxConfig config, IBoxService service, IBoxConverter
 
         private async Task<Dictionary<string, object>> UpdateMetadata(Uri hostUri, string id, List<BoxMetadataUpdate> updates, string scope, string template)
         {
+            foreach (BoxMetadataUpdate update in updates)
+            {
+                update.Path.ThrowIfNullOrWhiteSpace("Path");
+                update.Op.ThrowIfNullEnum("Op");
```
coolboy: Assume There is no " where T : class", we can use update.Op.ThrowIfNull("Op");

marek-vysoky:
I've removed where T : class from ThrowIfNull as you suggested

PR #238, opened 2016-11-09
Webhooks v2 review

cburnette:
added comments and integration tests for webhooks v2;
also added webhook validation method that depends on BouncyCastle-PCL library


boxcla:

PR #256, opened 2016-11-17
created a BoxCollectionMarkerBased class to handle the new paging sch…

cburnette:
…eme that uses next_marker field


boxcla:

PR #257, opened 2016-11-17
adds support for handling 429 errors (too_many_requests).  Will attem…

cburnette:
…pt to retry request maximum of 3 times observing RetryAfter header value.


boxcla:

Issue #267, opened 2017-01-01
can't add collaborator to files


Band-Aid:
client.CollaborationsManager.AddCollaborationAsync() only allows folders.
Need to allow to add collabs to files

PR #268, opened 2017-01-09
Add Single File Collab Support

coolboy:


boxcla:

coolboy:
fix for #267

cburnette:
@coolboy I think this works fine since it's a super class of BoxFolder

PR #276, opened 2017-01-19
fix so request is not retried for a 429 (rate limit) error on file up…

cburnette:
…loads (multipart request). The reason for this is that the stream has already been consumed by the time the error is received and there is not enough information about the stream to reset it.

Added a ResponseHeaders field to BoxException and set that value where appropriate (all response processing).


boxcla:

PR #290, opened 2017-03-02
Mupload p1

rpalaniappan:
- Added comments


boxcla:

Box.V2/Managers/BoxFilesManager.cs
```diff
@@ -328,7 +328,7 @@ public BoxFilesManager(IBoxConfig config, IBoxService service, IBoxConverter con
         /// Get a list of parts that were uploaded in a session.
         /// </summary>
         /// <param name="sessionPartsUri">The Url returned in the Create Session response.</param>
-        /// <returns></returns>
+        /// <returns>Returns a list of file part information uploaded so far in the session.</returns>
```
coolboy: Why not follow BoxCollection<T> pattern?

Ref:
public async Task<BoxCollection<BoxFileVersion>> ViewVersionsAsync(string id, List<string> fields = null)

PR #291, opened 2017-03-05
Mupload p1

rpalaniappan:
- Integration tests for uploading in a session
    - 1 to verify uploading and aborting works
    - 1 to verify uploading and committing works (and deletes the file after committing)
- Updated data type of file size to long from string
- Updated Abort/Delete session signature to take Uri instead of Id
- Updated comments


boxcla:

PR #296, opened 2017-03-09
Multiupload utility portable library

rpalaniappan:
- Utility function can be called using the portable library as well

### DIFF
--- a/Box.V2/Managers/BoxFilesManager.cs
+++ b/Box.V2/Managers/BoxFilesManager.cs
@@ -374,8 +374,9 @@ namespace Box.V2.Managers
         /// <param name="fileName">Name of the remote file name.</param>
         /// <param name="fileSize">Size of the file.</param>
         /// <param name="folderId">parent folder id.</param>
+        /// <param name="sha1HashFunction">Function delegate that returns message digest of the stream, formatted as specified by RFC 3230.</param>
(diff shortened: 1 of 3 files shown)


boxcla:

Box.V2/Managers/BoxFilesManager.cs
```diff
@@ -374,8 +374,9 @@ public BoxFilesManager(IBoxConfig config, IBoxService service, IBoxConverter con
         /// <param name="fileName">Name of the remote file name.</param>
         /// <param name="fileSize">Size of the file.</param>
         /// <param name="folderId">parent folder id.</param>
+        /// <param name="sha1HashFunction">Function delegate that returns message digest of the stream, formatted as specified by RFC 3230.</param>
         /// <returns></returns>
-        public async Task<bool> UploadUsingSessionAsync(Stream stream, string fileName, long fileSize, string folderId)
+        public async Task<bool> UploadUsingSessionAsync(Stream stream, string fileName, long fileSize, string folderId, Func<Stream,string> sha1HashFunction)
```
coolboy: why this returns bool instead of complete BoxFile object?

coolboy: why sha1HashFunction is on the public interface, could we hide it and auto select internally?

PR #300, opened 2017-03-29
Azure WebJob and Functions Fix for BoxJWTAuth

kendomen:
When running JWT in an Azure WebJob or Function, you will get an error:
System.Security.Cryptography.CryptographicException: The system cannot find the file specified.

The error message is misleading and you can see more details on it here:
https://blogs.msdn.microsoft.com/winsdk/2009/11/16/opps-system-security-cryptography-cryptographicexception-the-system-cannot-find-the-file-specified/

In order to fix this, I had to make changes to BoxJWTAuth to override the sealed class DotNetUtities so that I can add this line:

// This needs to be here to run as WebJob
csp.Flags = CspProviderFlags.UseMachineKeyStore;

### DIFF
--- a/Box.V2.JWTAuth/BoxJWTAuth.cs
+++ b/Box.V2.JWTAuth/BoxJWTAuth.cs
@@ -60,7 +60,6 @@ namespace Box.V2.JWTAuth
             // ------------------------------------------------------------------
             var rsa = ToRSA((RsaPrivateCrtKeyParameters)key.Private);   // use the local version to override the call to DotNetUtilities
 
-
 #if NETSTANDARD1_4    
             this.credentials = new SigningCredentials(new RsaSecurityKey(rsa), SecurityAlgorithms.RsaSha256);
 #else
--- a/Box.V2.JWTAuth/BoxJWTAuth.cs
+++ b/Box.V2.JWTAuth/BoxJWTAuth.cs
@@ -63,7 +63,7 @@ namespace Box.V2.JWTAuth
 #if NETSTANDARD1_4    
             this.credentials = new SigningCredentials(new RsaSecurityKey(rsa), SecurityAlgorithms.RsaSha256);
 #else
-            this.credentials = new SigningCredentials(new RsaSecurityKey(rsa), SecurityAlgorithms.RsaSha256Signature, SecurityAlgorithms.Sha256Digest);
+            this.credentials = new SigningCredentials(new RsaSecurityKey(rsa), Securiorithms.RsaSha256Signature, SecurityAlgorithms.Sha256Digest);
 #endif
         }
 
@@ -76,7 +76,7 @@ namespace Box.V2.JWTAuth
         //
         // I took the guts of the sealed class DotNetUtilities and overrode it here
         // -------------------------------------------------------------------------------------
-        public static RSA ToRSA(RsaPrivateCrtKeyParameters privKey)
+        private static RSA ToRSA(RsaPrivateCrtKeyParameters privKey)
         {
             return CreateRSAProvider(ToRSAParameters(privKey));
         }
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @kendomen has signed the CLA. Thanks for the pull request!

Box.V2.JWTAuth/BoxJWTAuth.cs
```diff
@@ -53,14 +54,77 @@ public BoxJWTAuth(IBoxConfig boxConfig)
             {
                 key = (AsymmetricCipherKeyPair)new PemReader(reader, pwf).ReadObject();
             }
-            var rsa = DotNetUtilities.ToRSA((RsaPrivateCrtKeyParameters)key.Private);
+            // ------------------------------------------------------------------
+            // This is the original call to BouncyCastle
+            // var rsa = DotNetUtilities.ToRSA((RsaPrivateCrtKeyParameters)key.Private);
+            // ------------------------------------------------------------------
+            var rsa = ToRSA((RsaPrivateCrtKeyParameters)key.Private);   // use the local version to override the call to DotNetUtilities
```
coolboy: Why not replacing the implementation on DotNetUtilities.ToRSA() instead of having a new one here?

Box.V2.JWTAuth/BoxJWTAuth.cs
```diff
@@ -53,14 +54,77 @@ public BoxJWTAuth(IBoxConfig boxConfig)
...
+            RSACryptoServiceProvider rsaCsp = new RSACryptoServiceProvider(csp);
+            rsaCsp.ImportParameters(rp);
+
+            return rsaCsp;
+        }
+
+        // ------------------------------------------------------------------------------------------------------------------
+        // Note: http://stackoverflow.com/questions/28370414/import-rsa-key-from-bouncycastle-sometimes-throws-bad-data
+        // ------------------------------------------------------------------------------------------------------------------
+        public static RSAParameters ToRSAParameters(RsaPrivateCrtKeyParameters privKey)
```
coolboy: we should not add a public method here

Box.V2.JWTAuth/BoxJWTAuth.cs
```diff
@@ -53,14 +54,77 @@ public BoxJWTAuth(IBoxConfig boxConfig)
...
+        // -------------------------------------------------------------------------------------
+        // When running JWT in an Azure WebJob or Function, you will get an error: 
+        //    "System.Security.Cryptography.CryptographicException: The system cannot find the file specified."
+        //
+        // The error message is misleading and you can see more info here:    
+        //     https://blogs.msdn.microsoft.com/winsdk/2009/11/16/opps-system-security-cryptography-cryptographicexception-the-system-cannot-find-the-file-specified/
+        //
+        // I took the guts of the sealed class DotNetUtilities and overrode it here
+        // -------------------------------------------------------------------------------------
+        public static RSA ToRSA(RsaPrivateCrtKeyParameters privKey)
```
coolboy: we should not add a public method here

coolboy:
Hi @kendomen 

Thank you for the PR, I just wonder why not just update the DotNetUtilities.ToRSA() instead of having a new one?

Box.V2.JWTAuth/BoxJWTAuth.cs
kendomen: I tried but it's a sealed private class.

Box.V2.JWTAuth/BoxJWTAuth.cs
kendomen: Ok, I'll make it private

Box.V2.JWTAuth/BoxJWTAuth.cs
kendomen: agreed

coolboy:
Hi @kendomen 

I have made another PR per your change, let me know if my PR works for you

https://github.com/box/box-windows-sdk-v2/pull/301/files

Issue #299, opened 2017-03-29
SDK should use api.box.com rather than app.box.com for the /token API


mself:
Calls to the OAuth /token API should use the `api.box.com domain` rather than `app.box.com`. See https://docs.box.com/reference#token

There are incorrect uses in `Box.V2.JWTAuth/BoxJWTAuth.cs` and `Box.V2/Auth/AuthRepository.cs`

coolboy:
fixed in #305

PR #305, opened 2017-04-05
SDK should use api.box.com for the /token API

coolboy:
#299


boxcla:

Issue #311, opened 2017-04-13
BoxJWTAuth constructor can give rise to System.NullReferenceException


Dave-Lowndes:
Debugging into the constructor code, the following:

`            AsymmetricCipherKeyPair key;
            using (var reader = new StringReader(this.boxConfig.JWTPrivateKey))
            {
                key = (AsymmetricCipherKeyPair)new PemReader(reader, pwf).ReadObject();
            }
`

results in key being null, so that the next line of code:

`            var rsa = DotNetUtilities.ToRSA((RsaPrivateCrtKeyParameters)key.Private);
`

throws a NullReferenceException.

I'm presuming the information I've been given that populates JWTPrivateKey may be wrong, but fundamentally it seems amiss that the constructor doesn't handle the key variable being null and throwing a more specific exception that could explain the issue precisely.

Any thoughts?

coolboy:
Hi @Dave-Lowndes ,

Thanks for the issue, will take look at it

PR #314, opened 2017-04-13
Throw exception on private key read error

coolboy:
#311

### DIFF
--- a/Box.V2.JWTAuth/BoxJWTAuth.cs
+++ b/Box.V2.JWTAuth/BoxJWTAuth.cs
@@ -1,6 +1,7 @@
 ﻿using Box.V2.Auth;
 using Box.V2.Config;
 using Box.V2.Converter;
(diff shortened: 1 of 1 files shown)


boxcla:

coolboy:
Want to look at it? @Dave-Lowndes

Dave-Lowndes:
That change is what I'd envisaged. I'm not too sure about the exception text message though - mostly because I don't know anything about PemReader or this code.

When I debugged the code, the PemReader constructor worked, but the result of ReadObject was null.
Rather than "Private key read error!", might it be more useful to say "Invalid private key" (though I don't know if the key I have is invalid or not)?

coolboy:
@Dave-Lowndes updated the message and merged

Thanks for the feedback!

Dave-Lowndes:
Thank you for such a speedy response.

PR #315, opened 2017-04-21
Multiput upload enhancement

coolboy:


boxcla:

cburnette:
@coolboy I assume this is still loading all the file into memory?

coolboy:
@cburnette this will load files into memory, but not all.

Internally it will load 5*8MB at the same time max and will release once the uploadpart is done

Per my testing (from 10MB -> 3GB files), the peak memory usage is around 250MB and avg. is around 100MB

coolboy:
@cburnette of course if you have the code sample showing me how to handle it better, I will see if I can update it.

PR #316, opened 2017-04-21
Update from string to object to support int/float update

coolboy:


boxcla:

PR #317, opened 2017-04-21
Use JWT with previously obtained UserToken

brunzefb:
Added a feature that allows using a userToken that was previously obtained by a call to the normal JWT Authentication mechanism.  This allows a web-service to do the login with the full credentials - client programs can then call the web service to retrieve a token and then create a UserClient with only the userToken. That means the client has less sensitive data, and only gets access to a token that times out. Drawback is that the clients have to manage refreshing the access token.

### DIFF
--- a/Box.V2.JWTAuth/BoxJWTAuth.cs
+++ b/Box.V2.JWTAuth/BoxJWTAuth.cs
@@ -39,11 +39,33 @@ namespace Box.V2.JWTAuth
 
         private readonly IBoxConfig boxConfig;
         private readonly SigningCredentials credentials;
-        /// <summary>
-        /// Constructor for JWT authentication
-        /// </summary>
-        /// <param name="boxConfig">Config contains information about client id, client secret, enterprise id, private key, private key password, public key id </param>
-        public BoxJWTAuth(IBoxConfig boxConfig)
+
+		/// <summary>
+		/// Constructor for JWT authentication - to be used if you already have a token. The second constructor
+		/// Parameter is just to differentiate it from the normal default constructor, value is not important.
+		/// </summary>
+		/// <code>
+		/// public void RefreshUserClient()
+		/// {
+		///	    var boxConfig = new BoxConfig("", "", "", "", "", "");
+		///     var boxJwt = new BoxJWTAuth(boxConfig, false);  
+		///     const string userToken = "USER_TOKEN_GOTTEN_FROM_NORMAL_JWTAUTH_WITH_FULL_CREDENTIALS";
+		///     UserClient = boxJwt.UserClient(userToken, null);
+		///     // the above UserClient will work for file calls, but will not support auto-refreshing of the tokens.
+		/// }
+		/// </code>
+		/// <param name="boxConfig">Config contains information about box config - but not necessarily the passwords.</param>
+		/// <param name="use">Not used</param>
+		public BoxJWTAuth(IBoxConfig boxConfig, bool use)
+	    {
+		    this.boxConfig = boxConfig;
+	    }
+
+	    /// <summary>
+		/// Constructor for JWT authentication
+		/// </summary>
+		/// <param name="boxConfig">Config contains information about client id, client secret, enterprise id, private key, private key password, public key id </param>
+		public BoxJWTAuth(IBoxConfig boxConfig)
         {
             this.boxConfig = boxConfig;
 


boxcla:
Hi @brunzefb, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

brunzefb:
I have signed the contributor's agreement.

Box.V2.JWTAuth/BoxJWTAuth.cs
```diff
@@ -39,11 +39,33 @@ public class BoxJWTAuth
...
-        public BoxJWTAuth(IBoxConfig boxConfig)
+
+		/// <summary>
+		/// Constructor for JWT authentication - to be used if you already have a token. The second constructor
+		/// Parameter is just to differentiate it from the normal default constructor, value is not important.
+		/// </summary>
+		/// <code>
+		/// public void RefreshUserClient()
+		/// {
+		///	    var boxConfig = new BoxConfig("", "", "", "", "", "");
```
coolboy: indent

Box.V2.JWTAuth/BoxJWTAuth.cs
```diff
@@ -39,11 +39,33 @@ public class BoxJWTAuth
...
+		///     var boxJwt = new BoxJWTAuth(boxConfig, false);  
+		///     const string userToken = "USER_TOKEN_GOTTEN_FROM_NORMAL_JWTAUTH_WITH_FULL_CREDENTIALS";
+		///     UserClient = boxJwt.UserClient(userToken, null);
+		///     // the above UserClient will work for file calls, but will not support auto-refreshing of the tokens.
+		/// }
+		/// </code>
+		/// <param name="boxConfig">Config contains information about box config - but not necessarily the passwords.</param>
+		/// <param name="use">Not used</param>
+		public BoxJWTAuth(IBoxConfig boxConfig, bool use)
```
coolboy: what does this bool use for?

brunzefb: It is to distinguish the constructor signature (explained in the header comment).  This is not ideal, but I don't see a better way of doing this which does not introduce a breaking change.

Box.V2.JWTAuth/BoxJWTAuth.cs
brunzefb: Do you guys use tabs or spaces on this project?

coolboy:
Hi @brunzefb,

If I understand your case correctly, what about below? Just need to replace accessToken with your token.

`            var auth = new OAuthSession(accessToken, "YOUR_REFRESH_TOKEN", 3600, "bearer");

            var config = new BoxConfig("YOUR_CLIENT_ID", "YOUR_CLIENT_ID", new Uri("http://boxsdk"));
            var client = new BoxClient(config, auth);

            var file = File.OpenRead(localFilePath);
            var fileRequest = new BoxFileRequest
            {
                Name = fileName,
                Parent = new BoxFolderRequest { Id = parentFolderId }
            };

            var bFile = await client.FilesManager.UploadUsingSessionAsync(file, fileName, parentFolderId);`

brunzefb:
@coolboy It would have been nice to discuss and get agreement with me before you closed the PR.  I don't think you fully understand what I am/we are trying to do. This PR has nothing to do with OAuth authentication, it is all about JWTAuth. 

I work in a team at a large company that uses the Box to create application accounts - UI-less accounts - only access is through custom apps that use the various Box SDK's. I am a very experienced C# developer and have experience doing uploads/downloads to other storage providers. Our team is creating one central web service that acts as an authentication broker -- that web service app has the full JWTAuth credentials required to obtain a token for a specific user.  The web service exposes an API that apps can log into and which allows retrieval of box JWT auth user tokens for specific users.  These JWT tokens are time-limited, and that's a good thing.  We have other applications that need to do uploads/downloads listing of files/folders on behalf of users.  Those apps make calls to the authentication broker, which hands them a fresh JWT token - these are using the full JWT functionality (with PEM certificate etc.) to get a user token. Using ONLY that JWT user token (and no other credentials), we have to create an instance of a BoxClient.  That is done by first creating a instance of the BoxJWTAuth class.  Now the constructor for that class takes all the credentials, and requires them to be valid, specifically the privateKey and JWTPrivateKey password.  If those two are null, empty or invalid, you get a nasty exception. But we need an instance of the BoxJWTAuth class without providing any valid credentials.  The code to get a UserClient (once you have a valid token) looks something like this then:

var boxConfig = new BoxConfig("", "", "", "", "", "");  // no passwords
var boxJwt = new BoxJWTAuth(boxConfig, false); // invoking the alternate constructor
const string userToken = "TOKEN"; // we have a valid JWT token from our web service
UserClient = boxJwt.UserClient(userToken, null); // and use it to get a valid UserClient.

I actually tried the above code with the changes in my original PR, and things(file/folder listing) worked fine.

Thinking about it some more, having a second constructor is kind of dumb - the BoxJWTAuth constructor can just check for null or empty strings and then not try to get credentials.  

I am going to make changes and submit another PR - let's have a dialog before you are rejecting the PR's from me, OK?  I am submitting these PR changes for a good reason - I think they will benefit others as well.

So no hard feelings from my end - just looking for a good discussion if you disagree on things.

Best,
Friedrich Brunzema

coolboy:
Hi @brunzefb 

Sure, and sorry to close the PR too early

Will review your new PR and use case on Monday to see what will be the right solution here

Thank you again for the response and feedback!

PR #319, opened 2017-04-22
Allow creation of BoxJWTAuth objects without JWTPrivateKey and password

brunzefb:
Added a feature that allows using a userToken that was previously obtained by a call to the normal JWT Authentication mechanism. This allows a web-service to do the login with the full credentials - client programs can then call the web service to retrieve a token and then create a UserClient with only the userToken. That means the client has less sensitive data, and only gets access to a token that times out. Drawback is that the clients have to manage refreshing the access token.


boxcla:
Verified that @brunzefb has signed the CLA. Thanks for the pull request!

brunzefb:
Created Issue  #320 to track

PR #330, opened 2017-05-12
Solution structure update part-1

coolboy:
2. Updated git ignore
3. Moved the some old examples to Legacy folder


boxcla:

Issue #297, opened 2017-03-28
When running JWT in an Azure WebJob or Function, you will get an error


kendomen:
When running JWT in an Azure WebJob or Function, you will get an error: 
           System.Security.Cryptography.CryptographicException: The system cannot find the file specified.
            The error message is misleading and you can see more details on it here:    
             https://blogs.msdn.microsoft.com/winsdk/2009/11/16/opps-system-security-cryptography-cryptographicexception-the-system-cannot-find-the-file-specified/

In order to fix this, I had to make changes to BoxJWTAuth:

 public BoxJWTAuth(IBoxConfig boxConfig)
        {
            this.boxConfig = boxConfig;

            // ------------------------------------------------------------------
            // Note:  When running JWT in an Azure WebJob or Function, you will get an error: 
            //           System.Security.Cryptography.CryptographicException: The system cannot find the file specified.
            //        The error message is misleading and you can see more details on it here:    
            //        https://blogs.msdn.microsoft.com/winsdk/2009/11/16/opps-system-security-cryptography-cryptographicexception-the-system-cannot-find-the-file-specified/
            // ------------------------------------------------------------------

            var pwf = new PEMPasswordFinder(this.boxConfig.JWTPrivateKeyPassword);
            AsymmetricCipherKeyPair key;

            using (var reader = new StringReader(this.boxConfig.JWTPrivateKey))
            {
                key = (AsymmetricCipherKeyPair)new PemReader(reader, pwf).ReadObject();
            }

            // ------------------------------------------------------------------
            // This is the original call to BouncyCastle
            // var rsa = DotNetUtilities.ToRSA((RsaPrivateCrtKeyParameters)key.Private);
            // ------------------------------------------------------------------
            var rsa = ToRSA((RsaPrivateCrtKeyParameters)key.Private);   // use the local overriden version instead!!!
            
#if NETSTANDARD1_4
            this.credentials = new SigningCredentials(new RsaSecurityKey(rsa), SecurityAlgorithms.RsaSha256);
#else
            this.credentials = new SigningCredentials(new RsaSecurityKey(rsa), SecurityAlgorithms.RsaSha256Signature, SecurityAlgorithms.Sha256Digest);
          
#endif
        }
       
        // -------------------------------- start KD Added 3-7-17 ------------------------------
        // Took the guts of the sealed class in DotNetUtilities and overrode it here
        // -------------------------------------------------------------------------------------
[...]

PR #334, opened 2017-05-17
Use our own RSA function to avoid problem in azure webfunction or oth…

coolboy:
…er platform under net core

#297


boxcla:

Issue #152, opened 2016-07-28
Unable to parse BoxCollection<BoxFile>


jimmyfakedev:
Hi,

I try to throw the BoxConflictException<T> exception when user try to upload an existing file but unfortunatelly, the Exception thrown is a BoxException type.

After debugging the project, I found the error in the BoxResponseExtensions.cs file at line 60:

`BoxConflictError<T> error = converter.Parse<BoxConflictError<T>>(response.ContentString);`

Am I doing something wrong or is it a real bug?

Regards,
Jimmy

coolboy:
hi @jimmyfakedev 

To be clear, you were expecting BoxConflictException for uploading a file with name already exist in that folder.

But the SDK gave you BoxException instead, correct?

PR #335, opened 2017-05-17
Handle the UploadFile conflict exception correctly

coolboy:
#152


boxcla:

PR #342, opened 2017-06-02
Updated retries to 5 after finding that retries actually happen - Whe…

brunzefb:
Updated retries to 5 after finding that retries actually do happen - When uploading parts, use as many threads as we have cores. R# cleaned and R# reformatted the UploadPartsInSessionAsync() and UploadUsingSessionAsync() methods.


boxcla:
Verified that @brunzefb has signed the CLA. Thanks for the pull request!

coolboy:
Thank you @brunzefb 

I am traveling to China right now, please expect some delay of merging this PR.

PR #346, opened 2017-06-13
Adding folder event source to converter (fixes #345)

### LINKED ISSUE #345: EnterpriseEventsAsync doesn't convert event_source if it's not a FileEventSource
Source is null for item_type "folder".  Is this just not implemented?

kendomen:
This fixes folder event sources getting converted from the response object.


boxcla:
Verified that @kendomen has signed the CLA. Thanks for the pull request!

PR #355, opened 2017-07-06
App User External Id Support

coolboy:
Missing JWTAuth cs files into Core


boxcla:

PR #400, opened 2017-09-25
Enterprise events datetime format

carycheng77:
Ran BoxEventsManagerTestIntegration with new datetime format and passed tests and retrieved same results from previous datetime format

### DIFF
--- a/Box.V2/Managers/BoxEventsManager.cs
+++ b/Box.V2/Managers/BoxEventsManager.cs
@@ -39,21 +39,9 @@ namespace Box.V2.Managers
                                                                         DateTime? createdAfter = null,
                                                                         DateTime? createdBefore = null)
         {
-            var createdAfterString = createdAfter.Value.ToString(createdAfter.Value.Kind == DateTimeKind.Utc 
-                ? Constants.RFC3339DateFormat_UTC 
-                : Constants.RFC3339DateFormat);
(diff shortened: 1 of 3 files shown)


boxcla:

Box.V2/Managers/BoxEventsManager.cs
```diff
@@ -39,7 +39,9 @@ public BoxEventsManager(IBoxConfig config, IBoxService service, IBoxConverter co
                                                                         DateTime? createdAfter = null,
                                                                         DateTime? createdBefore = null)
         {
-            var createdAfterString = createdAfter.HasValue ? createdAfter.Value.ToString(Constants.RFC3339DateFormat) : null;
+            var createdAfterString = createdAfter.Value.ToString(createdAfter.Value.Kind == DateTimeKind.Utc
```
coolboy: Where are the integration test? What about the `o` format?

ohadschn: More importantly, what about the proper conversion to UTC ? https://msdn.microsoft.com/en-us/library/system.datetime.touniversaltime(v=vs.110).aspx

ohadschn:
@carycheng77 @coolboy looks like you removed the `createdAfter.HasValue` check, this will result in `InvalidOperationException` if called with (the default) null value!

Also, why did you remove `WebUtility.UrlEncode`?

carycheng77:
@ohadschn That was a mistake on my part! 

I submitted a new pr with the null check put back in place and the url encoding

#403

PR #403, opened 2017-09-26
Fix eventsmanager mistake

carycheng77:
put back encoding if not null and added two additional test cases

1.) if one optional field for time is null
2.) if both optional fields for time is null


boxcla:

PR #408, opened 2017-10-18
fixed UploadUsingSessionInternal.GetNumberOfParts method.

rassy:
if totalSize is divisible by partSize, numberOfParts are counted 1 more.

### DIFF
--- a/Box.V2.Test.Integration/BoxFilesManagerTestIntegration.cs
+++ b/Box.V2.Test.Integration/BoxFilesManagerTestIntegration.cs
@@ -499,11 +499,10 @@ namespace Box.V2.Test.Integration
             {
                 if (partSize == 0)
                     throw new Exception("Part Size cannot be 0");
-                int numberOfParts = 1;
-                if (partSize != totalSize)
+                int numberOfParts = Convert.ToInt32(totalSize / partSize);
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @rassy, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

carycheng77:
Hey @rassy, 

We just tested three options and it seems to be okay
- FileSize < PartSize
- FileSize > PartSize
- FileSize == PartSize

And all three use cases seems to be returning the same value for both the original snippet as well as your snippet. Would love to gather more information regarding the issue you were experiencing and the FileSize and PartSize value you were using.

Thank you!

rassy:
Hi @carycheng77.

When I uploaded just 200MB(209,715,200byte),
an exception occurred.
=>System.FormatException: The format of value 'bytes 209715200-209715199/209715200' is invalid.
(Box returned [size=8388608] by response header. = partSize)

I think GetNumberOfParts must return 25, but return 26 at this case.

Best regards.

carycheng77:
Ah! Good catch @rassy  

I tested it again and there is indeed an issue. Thank you for the PR. We updated our code and added test cases surrounding this as well.

PR #410, opened 2017-10-19
fix for get number of parts

carycheng77:
Hey Yang, I broke the test method out into a separate utility method test because I didn't want to clog UploadFileInSession_AbortRequest_FileNotCommmited() with all the part size number tests. Let me know what you think!


boxcla:

PR #415, opened 2017-11-09
Metadata template id update

carycheng77:
Thanks Yang!


boxcla:

PR #425, opened 2017-12-20
added two test cases and minor change to the documentation for single…

carycheng77:
Hey Yang, very minor change to this in the documentation because we support files on top of folders now for file collaboration, let me know if this is necessary, thanks! 

I saw you had a test to add user collab on a single file by login so I added another one for adding single file collab with user ID and group ID. 

Thanks!


boxcla:

PR #427, opened 2018-01-03
removed required file name for restore file from trash. Added tests f…

carycheng:
…or restore file and restore folder from trash


boxcla:

PR #429, opened 2018-01-04
Modified 'Status' property of the BoxRepresentationStatus class to overcome deserialization issue

PraveenKariyapperuma:
I was working with representations endpoint and found that the status of the representation is always null. I Inspected the endpoint with fiddler and realized that the response contains a property called 'state' where the BoxRepresentationStatus class has defined it as 'Status'. So i went ahead and renamed property name to 'State' and the Json property name to 'state' accordingly.

### DIFF
--- a/Box.V2/Models/BoxRepresentationStatus.cs
+++ b/Box.V2/Models/BoxRepresentationStatus.cs
@@ -4,12 +4,12 @@ namespace Box.V2.Models
 {
     public class BoxRepresentationStatus
     {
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @PraveenKariyapperuma, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

carycheng:
@PraveenKariyapperuma, oversight on my part, really appreciate the pr!

PR #431, opened 2018-01-05
Add docs for App Token Auth

### LINKED ISSUE #430: How to use the SDK with an "App Token (Server Authentication)?
We're writing an app that needs to be able to upload files to Box so that we can use the "New Box View" preview functionality. I've created an app in the Box dev console, and specified "App Token (Server Authentication)" as the authentication type, as specified by [the migration docs](https://developer.box.com/docs/getting-started-with-new-box-view). 

Looking at the SDK readme for .NET and I'm having a heck of a time trying to decipher how the "App Token" style of authentication jives with what's in the readme. In the readme I see terminology like "Client Id", "Client Secret", "Developer Token," etc. But in the new app I've created, the terminology is "API Key", "Primary Access Token", "Secondary Access Token" (which I understand is just a fallback that works like the primary). 

So when I see this example of how to setup a client, I don't really understand how to utilize the key/token that has been generated:

```
var config = new BoxConfig(<Client_Id>, <Client_Secret>, new Uri("http://localhost"));
var session = new OAuthSession(<Developer_Token>, "NOT_NEEDED", 3600, "bearer");
client = new BoxClient(config, session);
```

Does anyone have an example of how to use the SDK with the App Token (Server Authentication)?

mattwiller:
Fixes #430

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,20 @@ Alternatively, a completely custom OAuth2 authentication process can be used in 
 OAuthSession session = // Create session from custom implementation
 var client = new BoxClient(config, session);
 ```
+
+#### Using With App Token Auth (Box View)
+
+Box View uses App Token Auth, where the app generated long-lived access tokens
+that are used directly in place of any authentication calls to the API.  To use
+this method of aythentication, simply create a client with only the API key and
+access token provided:
+
+```c#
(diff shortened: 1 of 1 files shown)


boxcla:

PR #433, opened 2018-01-10
Help build tools find Box.V2.Core nuspec file

mattwiller:
For some reason, the `dotnet` tool we use for release won't use the values in the nuspec file unless it's explicitly mentioned in the csproj file.  Adding this ensures that the package metadata for the Box.V2.Core project will be present in the nuget package.


boxcla:

PR #452, opened 2018-02-26
Exponential backoff

carycheng:


boxcla:
Hi @carycheng, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

Box.V2/Utility/ExponentialBackoff.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
mattwiller: Do you need all of these?  It doesn't look like you use many different classes below.

Box.V2/Request/HttpRequestHandler.cs
```diff
@@ -82,12 +84,12 @@ public class HttpRequestHandler : IRequestHandler
...
                         ||
                         (response.StatusCode == HttpStatusCode.Accepted && retryAfterHeader != null)) 
-                        && numRetries-- > 0)
+                        && retryCounter++ < 3)
                     {
                         TimeSpan delay = TimeSpan.FromSeconds(2);
                         if (retryAfterHeader.Delta.HasValue)
                         {
-                            delay = retryAfterHeader.Delta.Value;
+                           delay = expBackoff.GetRetryTimeout(retryCounter, retryAfterHeader.Delta.Value);
```
mattwiller: I worry a bit about using both the Retry-After header and the exponential backoff — what if in the future the Retry-After header starts giving higher values?  Then the customer's backoff time will go up really fast.

Box.V2/Request/HttpRequestHandler.cs
```diff
@@ -82,13 +84,9 @@ public class HttpRequestHandler : IRequestHandler
                         ((response.StatusCode == TooManyRequests && !isMultiPartRequest)
                         ||
                         (response.StatusCode == HttpStatusCode.Accepted && retryAfterHeader != null)) 
-                        && numRetries-- > 0)
+                        && retryCounter++ < 3)
```
mattwiller: Can we pull this number into a constant at the top of the file and update its value to 5 to be in line with the other SDKs?  You should also update the test for the backoff counter to test values up to 5.

PR #456, opened 2018-03-13
Add JWT retry with new jti claim and updated exp time

mattwiller:


boxcla:

mattwiller:
@coolboy This is a work in progress, but I'd love to get feedback from you on the direction of this implementation before I go much further.  Does this seem like the right way to retry in cases where the exp claim was invalid, e.g. due to clock drift?

Also, how would I test something like this?

PR #458, opened 2018-03-16
Restrict ThrowIfNull check to reference and Nullable types

### LINKED ISSUE #375: Nit-picking point on BoxExtensions ThrowIfNull
Analysing the Box source with PVS Studio identifies this minor point in BoxExtensions.cs L18:

"V3111 Checking value of 'param' for null will always return false when generic type is instantiated with a value type."

Their [recommendation ](https://www.viva64.com/en/w/V3111/print/)is to either constrain the method to handle only reference types, or test against the type's default value.

mattwiller:
Fixes #375


boxcla:

PR #459, opened 2018-03-16
Fix parsing for user tracking codes

### LINKED ISSUE #424: tracking_codes field not properly parsed on BoxUser or BoxUserRequest
The tracking_codes field is currently a `string[]` and the API is returning a complex type with `type`, `name`, and `value` fields. The JsonConverter is failing when trying to parse this field on a BoxUser. I opened a pull request to address the issue #423.

mattwiller:
Moving the original change by @allenmichael to a feature branch in the SDK proper so CI tests pass.  Fixes #424

### DIFF
--- a/Box.V2.Test/BoxUsersManagerTest.cs
+++ b/Box.V2.Test/BoxUsersManagerTest.cs
@@ -43,6 +43,48 @@ namespace Box.V2.Test
         }
 
         [TestMethod]
+        public async Task GetUserInformation_TrackingCodes()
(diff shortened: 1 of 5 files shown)


boxcla:

mattwiller:
@carycheng Before we merge this, we should evaluate which approach to standardize on — the dedicated Tracking Code class like in this PR or the Map<String, String> sort of thing that just landed in the Java SDK.

PR #464, opened 2018-03-23
Fix protocol not supported error in .NET Core on MacOS

mattwiller:
Setting both `Tlsv11` and `Tlsv12` causes an error in .NET Core on MacOS — switching to just one or the other appears to resolve the issue.  Since TLSv1.2 is the preferred version, we will set that one to maximize security going forward.


boxcla:

PR #465, opened 2018-03-27
Fix event source subclass property overrides

### LINKED ISSUE #409: Box.V2.Models.BoxFileEventSource inherits from BoxEntity but hides all of it's inherited fields.
This creates the problem that when you receive an Admin Event from a File source you can't find a reference to the source through the interface.  This can be worked around by explicitly casting, but the child class should never hide the parent interface. (The use of the word never almost feels like cursing, but this type of issue leads to some really strange and hard to trace bugs.)  If the desire is to override the base class implementation then mark the base class properties virtual and mark the child class properties with override so that the end user will get the correct implementation no matter the interface used to reach the object. 

I would provide a pull request but in reading your CLA I can't speak on behalf of my time for my employer.  

Kindest Regards,
Dan Blair

mattwiller:
Fixes #409

### DIFF
--- a/Box.V2/Models/BoxEntity.cs
+++ b/Box.V2/Models/BoxEntity.cs
@@ -15,12 +15,12 @@ namespace Box.V2.Models
         /// The item’s ID
         /// </summary>
         [JsonProperty(PropertyName = FieldId)]
-        public string Id { get; private set; }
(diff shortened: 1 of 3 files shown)


boxcla:

PR #479, opened 2018-04-24
Catch exceptions when attempting to set TLSv1.2 protocol

### LINKED ISSUE #476: Client initialization fails in Xamarin iOS and Android.
Hi!
I've been using the client for a while and after upgrading to the latest version it crashes.
HttpClientHandler.SslProtocols property throws in those platforms with a NotImplementedException. The exception is actually raised by the property getter because of the |= assignment.

Here's the offending line:

https://github.com/box/box-windows-sdk-v2/blob/816c66770f9438a1faf1dcd71de0e4fcd04890e2/Box.V2/Request/HttpRequestHandler.cs#L189

Cheers,
Andrés

mattwiller:
Fixes #476


boxcla:

PR #490, opened 2018-06-12
Allow setting can_non_owners_invite flag on folders

mattwiller:


boxcla:

PR #518, opened 2019-01-02
added methods to 3 Managers - see comments

ajweber:
FilesManager and FoldersManager ADDED:
public async Task<bool> AddFileToFavoritesAsync(string Id, string myFavoritesId);
public async Task<bool> RemoveFileFromFavoritesAsync(string Id);

FilesManager, FoldersManager, and WebLinksManager ADDED:
public async Task<IBoxResponse<BoxFile>> RefreshInformationAsync(string id, string eTag, List<string> fields = null);

Constants needed If-None-Match http header constant.

### DIFF
--- a/Box.V2/Config/Constants.cs
+++ b/Box.V2/Config/Constants.cs
@@ -241,7 +241,8 @@ namespace Box.V2.Config
             public const string Pending = "pending";
 
             public const string IfMatch = "If-Match";
-
+            public const string IfNoneMatch = "If-None-Match";
+            
             /*** Values for specifically representations endpoint ***/
             public const string XRepHints = "x-rep-hints";
             public const string SetContentDispositionType = "set_content_disposition_type";
--- a/Box.V2/Managers/BoxFilesManager.cs
+++ b/Box.V2/Managers/BoxFilesManager.cs
@@ -47,6 +47,73 @@ namespace Box.V2.Managers
         }
 
         /// <summary>
+        /// Gets a file object representation of the provided file Id IF IT'S CHANGED.
+        /// </summary>
+        /// <param name="id">Id of file information to retrieve</param>
+        /// <param name="eTag">Current eTag (to check against the Box service</param>
+        /// <param name="fields">The fields to return as part of the response if changed</param>
+        /// <returns></returns>
+        public async Task<IBoxResponse<BoxFile>> RefreshInformationAsync(string id, string eTag, List<string> fields = null)
+        {
+            id.ThrowIfNullOrWhiteSpace("id");
+            eTag.ThrowIfNullOrWhiteSpace("eTag");
+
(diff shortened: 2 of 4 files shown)


boxcla:
Verified that @ajweber has signed the CLA. Thanks for the pull request!

carycheng:
Hey @ajweber, 

Thank you for the pull request! 

There are a few things to note here, the diffs for files changed shows that the entire file for all three managers was replaced and that makes it hard to see the changes you did.

If you are going to resubmit a PR then it might be a good idea to break the favorites functionality for files and folders into one PR and weblinks into a separate PR!

Also for each new function written we required a unit test to be written as well!

For examples please refer to [here](https://github.com/box/box-windows-sdk-v2/blob/master/Box.V2.Test/BoxFilesManagerTest.cs#L56)

We really appreciate it and please let us know if any questions may arise!

ajweber:
Sorry, Cary, I didn't write github's diff util. If you take a glance at the files, you can see that I added 3 new methods to the 2 "managers" and one to the other.

The only other changes are making sure that the "using System.Net" is in each (to ref the standard HTTP codes) and one line in the constants file for the If-Not-Match header.

I think these are extremely useful and completely transparent to existing code. 

I can probably devise a test for each, but I don't know what to tell you about github's diff. I was offering these edits as a way to improve the SDK...not to create a lot of work in addition to my day job. 🙂

whiggs:
Completely agree with @ajweber .  People who are gracious enough to take the time to improve a project and submit those improvements likely do so when they have additional responsibilities (like jobs) which consume the majority of their time.  However, any trivialities that repo owners require (like fixing invalid, broken unit testing, performing the trivial tasks that are demanded be performed before request can be merged, etc.) are the RESPONSIBILITY OF THE REPO OWNER, NOT THE PULL REQUESTER!!!  In most cases, these projects (like this one, for example) are the products of a business, where this argument is even more valid.  Not that box is the only company to adopt this flawed, short sighted approach (shout out to Powershellget project for stupid is as stupid does). but, in the end, if the goal is to improve your product, you shoot yourselves in the foot by using these flawed, restrictive, and just stupid policies.

mattwiller:
At the end of the day, every piece of code in this repository is our team's responsibility, and we're not trying to force anyone outside of Box to do any work they don't want to do — after all, it's not their job!  However, any code that's being merged into the SDK needs to adhere to the standards of the repository.  If the author requests that we do the additional work in order to get a pull request into a mergeable state, we can certainly do that for them!  It just may not happen right away, depending on what other work our team is doing at the time.

@whiggs That said, this is not an appropriate forum for derogatory statements about this project or any other.  As with all of our open-source projects, we expect everyone participating in this repository to abide by the [Box OSS Code of Conduct](http://opensource.box.com/code-of-conduct/) and be respectful and considerate.

PR #520, opened 2019-01-10
remove reference to "beta" chunked upload

jasonpan12:


boxcla:
Verified that @jasonpan12 has signed the CLA. Thanks for the pull request!

PR #521, opened 2019-01-11
Conditionally serialize unshared_at property

### LINKED ISSUE #512: Cannot change shared link back to never expires once an expiration date is set
- [ ] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
Once a file's shared link has a `DateTime` set on its expiration date, cannot set it back to `null`. As a result, we can't set the shared link back to never expires.

### Versions Used
.NET SDK: 3.9.0
.NET Framework/Core: .net core 1.1

### Steps to Reproduce
1. Create shared link on a file
2. Set expiration date on the shared link
3. Set expiration date back to never expire (null)
4. Expiration date is still the same from Step 2

**SDK Code (doesn't work)**
```
var requestParams = new BoxFileRequest()
    {
        Id = sourceDocumentId,
        SharedLink = new BoxSharedLinkRequest()
        {
            UnsharedAt = null
        }
    };
 
boxFile = await _boxServiceClient.BoxAuthentication(boxUserId)
    .FilesManager
    .UpdateInformationAsync(requestParams)
    .ConfigureAwait(false);
```

**API Call (works)**
```
https://api.box.com/2.0/files/:fileId?fields=shared_link
{
    "shared_link": {
        "unshared_at": null
    }
}
```

No Error message.

[sdk-docs]: ./README.md

mattwiller:
Use conditional serialization to ensure that the shared link `unshared_at` property is correctly sent to the API:

- A `null` value should be sent to the API
- The `null` value should not be sent if the user did not actively set it

Fixes #512

### DIFF
--- a/Box.V2.Test/BoxFilesManagerTest.cs
+++ b/Box.V2.Test/BoxFilesManagerTest.cs
@@ -1,4 +1,4 @@
-﻿using Box.V2.Managers;
+using Box.V2.Managers;
 using Box.V2.Models;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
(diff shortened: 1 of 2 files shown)


boxcla:

PR #527, opened 2019-01-23
Add support for getting User Avatar

### LINKED ISSUE #525: Any plan to support the new Get User Avatar endpoint?
According to https://developer.box.com/docs/api-changelog-2017 avatar_url only returns a static image now.

Is there any plan to support the new Get User Avatar on .NET's SDK?

Thanks.

Kind regards,
Raymond

mattwiller:
Added `client.UsersManager.GetUserAvatar(string userId)`, which returns a `Stream` over the bytes of the user's avatar image.

Fixes #525


boxcla:

PR #531, opened 2019-02-01
Add correct method for retrieving file collaborations

### LINKED ISSUE #524: "GetCollaborationsAsync" method of the "BoxFilesManager" class does not return number of collaborations
### Description of the Issue
So I have found that using the "GetCollaborationsAsync" method of the "BoxFilesManager" class does not return the correct number of collaborations for a file, even though it does list the collaboration entries.  I am creating a powershell module for box which utilizes the box windows sdk and found this while testing my module's functions. Consider the below code:

```powershell
function get-boxfolderitems
{
<#
This function returns the "result" property of the variable which stores the result of using "GetFolderItemsAsync" method in "BoxFoldersManager" class
#>
	[CmdletBinding(DefaultParameterSetName = "auto")]
	param (
		[string]$folderid = "0",
		[Parameter(ParameterSetName = 'noauto')]
		[int]$offset = 0,
		[Parameter(ParameterSetName = 'noauto')]
		[int]$limit = 1000,
		[Parameter(ParameterSetName = 'auto')]
		[Switch]$autopage,
		[System.String[]]$fields
	)
	If ($PSCmdlet.ParameterSetName -like "auto")
	{
		$foldlist = $adminclient.FoldersManager.GetFolderItemsAsync($folderid, 1000, 0, $PSBoundParameters.fields, $true)
	}
	Else
	{
		$foldlist = $adminclient.FoldersManager.GetFolderItemsAsync($folderid, $limit, $offset, $PSBoundParameters.fields, $false)
	}
	$foldlist.Wait()
	If ($foldlist.IsFaulted -eq $true)
	{
		Write-error $foldlist.Exception.InnerException
		return
	}
	Else
	{
		return $foldlist.Result
	}
}

$tt = get-boxfolderitems -autopage
# Takes the box folder items, filters out everything but the files, and passes that to the "GetCollaborationsAsync" method of the "BoxFilesManager" class

$tt.Entries | Where-object {$_.Type -like "file"} | % {($adminclient.FilesManager.GetCollaborationsAsync($_.id)).Result}

```

You should also assume that there is a file in "$tt.entries" that has a collaboration.  In any event here is the result that we see when running the above code:

image

As you can see, even though a collaboration is detected and listed in the "entries" property, total count remains 0 for some reason, when it should be 1.

### Versions Used
.NET SDK: 3.11.0

### Steps to Reproduce
Please see above.  I was pretty clear about how to reproduce the issue at the very beginning and would rather not repeat myself.

No error message, just incorrect return value.

[sdk-docs]: ./README.md

mattwiller:
Fixes #524

### DIFF
--- a/Box.V2.Test.Integration/BoxFilesManagerTestIntegration.cs
+++ b/Box.V2.Test.Integration/BoxFilesManagerTestIntegration.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading.Tasks;
 using System.Collections.Generic;
 using System.IO;
--- a/Box.V2.Test/BoxFilesManagerTest.cs
+++ b/Box.V2.Test/BoxFilesManagerTest.cs
@@ -9,8 +9,6 @@ using System.Net.Http.Headers;
 using System.IO;
 using Newtonsoft.Json;
 using System.Text;
-using Box.V2.Config;
-using Box.V2.Auth;
 
 namespace Box.V2.Test
 {
--- a/Box.V2.Test/BoxFilesManagerTest.cs
+++ b/Box.V2.Test/BoxFilesManagerTest.cs
@@ -9,6 +9,8 @@ using System.Net.Http.Headers;
 using System.IO;
 using Newtonsoft.Json;
 using System.Text;
+using Box.V2.Config;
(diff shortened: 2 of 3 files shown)


boxcla:

PR #534, opened 2019-02-13
Add missing fields to User model

### LINKED ISSUE #533: Users API endpoint returns hostname field but it's not exposed on the BoxUser model
### Problem
According to the Box API documentation, the Users API endpoint returns a hostname field for _the root (protocol, subdomain, domain) of any links that need to be generated for the user_. This field is returned by the API but not parsed out of the response and included as a property of the BoxUser model.

### Versions Used
.NET Standard & .NET Framework: v3.11.0

### Description of the Issue
My application needs to know the hostname of the Box app of any given user. For example, it appears to me that the User object's hostname field is the best way to generate an absolute URL for a file in Box. Using fiddler you can see the hostname is in the API response from BoxUsersManager.GetCurrentUserInformationAsync if you include the hostname field in the initial request.

I modified the source code locally and propose the following changes after confirming they work in my local dev: modify the BoxUser class by adding a public const string for the hostname field and adding a public string Hostname property to expose it from the request. I can submit a pull request if approved.

mattwiller:
Fixes #533

### DIFF
--- a/Box.V2.Test/BoxUsersManagerTest.cs
+++ b/Box.V2.Test/BoxUsersManagerTest.cs
@@ -86,6 +86,41 @@ namespace Box.V2.Test
         }
 
         [TestMethod]
+        public async Task GetUserInformation_ExtraFields()
+        {
+            /*** Arrange ***/
+            string responseString = @"{
+                ""type"": ""user"",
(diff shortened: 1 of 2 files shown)


boxcla:

PR #536, opened 2019-02-20
Override Id property in EventSource classes

### LINKED ISSUE #535: BoxGroupEventSource Issue?
- [x] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue

I believe I found a bug in BoxGroupEventSource. The Id field is not declared with override to override the Id field in BoxEntity. This is inconsistent with how BoxFolderEventSource and BoxFileEventSource declarations which which both override the Id field. This is showing up when we process events and causing null exceptions at times. Some events have the BoxEntity Id field set while other events have the Id field from BoxGroupEventSource set. You can access both fields based on how the object is cast. Can you verify if this is a bug or "feature". I'm thinking bug.

### Versions Used
.NET SDK: 2.3.12 

### Steps to Reproduce
Since this is a design question the only steps are to look at the above mentioned source code.

Id is sometimes blank if you don't cast the object to the appropriate type.

[sdk-docs]: ./README.md

mattwiller:
Many of the EventSource classes defined an `Id` property which shadows the base `BoxEntity` property of the same name.  Without marking the subclass property as `override`, the data exposed would vary depending on how the object was cast.  Marking the property as `override` should ensure that the correct data is always exposed.

Fixes #535

### DIFF
--- a/Box.V2/Models/BoxGroupEventSource.cs
+++ b/Box.V2/Models/BoxGroupEventSource.cs
@@ -1,4 +1,4 @@
-﻿using Newtonsoft.Json;
+using Newtonsoft.Json;
 
 namespace Box.V2.Models
 {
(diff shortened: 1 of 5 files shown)


boxcla:

PR #537, opened 2019-02-23
Fix deserialization of translated task assignment status

### LINKED ISSUE #529: TaskAssignment request fail by box.V2
I tried the following sdk: 
https://github.com/box/box-windows-sdk-v2/blob/5800ee9aa621493d6d23d8628b2bf5b20057309a/docs/tasks.md#assign-task

I execute CreateTaskAssignmentAsync and it returns error, 

> Error converting value "未完了" to type 'System.Nullable`1[Box.V2.Models.ResolutionStateType]'. Path >'resolution_state', line 1, position 564. 
>Newtonsoft.Json.JsonSerializationException: Error converting value "未完了" to type >'System.Nullable`1[Box.V2.Models.ResolutionStateType]'. Path 'resolution_state', line 1, position 564. ---> >System.ArgumentException: Requested value '未完了' was not found. 

"未完了" means "Incomplete". 
but I can find on box web that the task is assigned to user without problem.

So this may be a bug with box.V2 and please check this.

Your support team said "This seems to be localization-related when SDK is trying to parse in the Task status."
So please fix this to use in Japanese environment.

Thanks,
Tomoyuki Kumaoka

mattwiller:
The `resolution_state` field of a task assignment can inadvertently contain localized text, which breaks the automatic deserialization into an enum and will throw as soon as the object is created.  The current enum solution seems too brittle, so it has been deprecated in favor of two new `string` properties: `.StatusTranslated` for the localized text and `.Status` for the new `status` field that will only contain the enum string values.  The deprecated `.ResolutionState` property has been re-implemented to convert to enum from the `status` field, which should prevent it from throwing on instantiation while maintaining backward compatibility.

Fixes #529

### DIFF
--- a/Box.V2/Models/BoxTaskAssignment.cs
+++ b/Box.V2/Models/BoxTaskAssignment.cs
@@ -58,7 +58,7 @@ namespace Box.V2.Models
         /// <summary>
         /// Gets the state of the resolution.
         /// </summary>
-        [Obsolete("This field is deprecated, and may not work consistently.  Use Status or StatusTranslated instead.")]
+        [Obsolete("This field is deprecated, and may not work consistently.  Use Status or LocalizedStatus instead.")]
         public ResolutionStateType? ResolutionState {
             get
             {
@@ -73,10 +73,10 @@ namespace Box.V2.Models
         public string Status { get; private set; }
 
         /// <summary>
-        /// Gets the translated/human-readable resolution status of the task assignment.
+        /// Gets the localized/human-readable resolution status of the task assignment.
         /// </summary>
         [JsonProperty(PropertyName = FieldResolutionState)]
-        public string StatusTranslated { get; private set; }
+        public string LocalizedStatus { get; private set; }
 
         /// <summary>
         ///Gets user assigned by.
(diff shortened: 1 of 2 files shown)


boxcla:

Box.V2/Models/BoxTaskAssignment.cs
```diff
@@ -57,9 +58,25 @@ public class BoxTaskAssignment : BoxEntity
...
+        [JsonProperty(PropertyName = FieldStatus)]
+        public string Status { get; private set; }
+
+        /// <summary>
+        /// Gets the translated/human-readable resolution status of the task assignment.
+        /// </summary>
         [JsonProperty(PropertyName = FieldResolutionState)]
-        [JsonConverter(typeof(StringEnumConverter))]
-        public ResolutionStateType? ResolutionState { get; private set; }
+        public string StatusTranslated { get; private set; }
```
nidoran: Use "Localization" instead of "Translation" to sound more technical, so either `StatusLocalized` or `LocalizedStatus`. Personally, I would like the new fields to be `CanonicalStatus` and `LocalizedStatus`, but I can see the advantage of having API response fields consistently map to SDK object fields (ie I'm fine with `Status` and `LocalizedStatus`).

PR #554, opened 2019-04-09
Encode query parameters

### LINKED ISSUE #543: User search with special character in the email
- [X] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
We are using GetEnterpriseUsersAsync(..) (.net SDK) to search for a given user to see if the user exists in our instance. This call takes many arguments but we only pass the “filterTerm” and “fields”.

Some of the email addresses have special characters in them.  With the “filterTerm” parameter, we are seeing that if we pass an email address with a “+” in it , it is not matching anything and does not return  a result.  If you replace + with %2B it returns the correct result match.  I have tested this in the Java SDK and CLI v2 and the search call performs differently than .NET.  The "+" does not need to be translated in those cases and it searches as expected.  

### Versions Used
.NET SDK: 3.4.2 

### Steps to Reproduce
client.UsersManager.GetEnterpriseUsersAsync(filterTerm: “jtrujillo+test@box.com”, fields:null)

mattwiller:
Fixes #543


boxcla:

PR #556, opened 2019-04-12
Retry on 5xx errors

mattwiller:


boxcla:

Box.V2/Request/HttpRequestHandler.cs
```diff
@@ -81,8 +81,14 @@ public async Task<IBoxResponse<T>> ExecuteAsync<T>(IBoxRequest request)
                     // If we get a 429 error code and this is not a multi part request (meaning a file upload, which cannot be retried
```
PJSimon: Suggest changing "a 429" to "an".

PR #559, opened 2019-04-26
Tag code samples for usage in developer docs

cbetta:
I've tagged the majority of code samples to an endpoint for usage in the new developer documentation.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -36,6 +36,7 @@ token from your application's
 
 The following example creates an API client with a developer token:
 
+<!-- sample get_authorize -->
 ```c#
 var config = new BoxConfig("YOUR_CLIENT_ID", "YOUR_CLIENT_SECRET", new Uri("http://localhost"));
 var session = new OAuthSession("YOUR_DEVELOPER_TOKEN", "N/A", 3600, "bearer");
--- a/docs/collaborations.md
+++ b/docs/collaborations.md
@@ -26,6 +26,7 @@ A collaboration can be added for an existing user by calling
 The `Role` field of the `collaborationRequest` parameter determines what permissions the collaborator will have on the
 folder or file. 
 
+<!-- sample post_collaborations -->
 ```c#
 // collaborate folder 11111 with user 22222
 BoxCollaborationRequest requestParams = new BoxCollaborationRequest()
@@ -54,6 +55,7 @@ A collaboration can be edited by calling
 `CollaborationsManager.EditCollaborationAsync(BoxCollaborationRequest collaborationRequest, IEnumerable<string> fields = null)`
 with the fields to be updated.  For example, to change the role of a collaboration:
 
+<!-- sample put_collaborations_id -->
 ```c#
 BoxCollaborationRequest requestParams = new BoxCollaborationRequest()
 {
@@ -69,6 +71,7 @@ Remove a Collaboration
 A collaboration can be removed by calling `CollaborationsManager.RemoveCollaborationAsync(string id)`.
 This will generally remove the user or group's access to the collaborated item.
 
+<!-- sample delete_collaborations_id -->
 ```c#
 await client.CollaborationsManager.RemoveCollaborationAsync(id: "12345");
 ```
@@ -80,6 +83,7 @@ To get information about a specific collaboration record, call
 `CollaborationsManager.GetCollaborationAsync(string id, IEnumerable<string> fields = null)` with the
 ID of the collaboration.
 
+<!-- get delete_collaborations_id -->
 ```c#
 BoxCollaboration collab = await client.CollaborationsManager.GetCollaborationAsync(id: "22222");
 ```
@@ -90,6 +94,7 @@ Get the Collaborations on a Folder
 You can get all of the collaborations on a folder by calling
 `FoldersManager.GetCollaborationsAsync(string id, IEnumerable<string> fields = null)` with the ID of the folder.
 
+<!-- sample delete_collaborations_id -->
 ```c#
 string folderId = "11111";
 BoxCollection<BoxCollaboration> collaborations = await client.FoldersManager
@@ -115,6 +120,7 @@ Get Pending Collaborations
 A collection of all the user's pending collaborations can be retrieved with
 `CollaborationsManager.GetPendingCollaborationAsync(IEnumerable<string> fields = null)`.
 
+<!-- sample get_collaborations -->
 ```c#
 BoxCollection<BoxCollaboration> pendingCollabs = await CollaborationsManager
     .GetPendingCollaborationAsync();
--- a/docs/collaborations.md
+++ b/docs/collaborations.md
@@ -83,7 +83,7 @@ To get information about a specific collaboration record, call
 `CollaborationsManager.GetCollaborationAsync(string id, IEnumerable<string> fields = null)` with the
 ID of the collaboration.
 
-<!-- get delete_collaborations_id -->
+<!-- sample get_collaborations_id -->
 ```c#
 BoxCollaboration collab = await client.CollaborationsManager.GetCollaborationAsync(id: "22222");
 ```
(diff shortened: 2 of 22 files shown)


boxcla:
Verified that @cbetta has signed the CLA. Thanks for the pull request!

docs/authentication.md
```diff
@@ -36,6 +36,7 @@ token from your application's
 
 The following example creates an API client with a developer token:
 
+<!-- sample get_authorize -->
```
mattwiller: I'm not sure this is really the right sample for that endpoint; this code doesn't make any API calls, but rather instantiates a client with a developer token

docs/collaborations.md
```diff
@@ -80,6 +83,7 @@ To get information about a specific collaboration record, call
 `CollaborationsManager.GetCollaborationAsync(string id, IEnumerable<string> fields = null)` with the
 ID of the collaboration.
 
+<!-- get delete_collaborations_id -->
```
mattwiller: Should be `get_collaborations_id`

docs/collaborations.md
```diff
@@ -90,6 +94,7 @@ Get the Collaborations on a Folder
 You can get all of the collaborations on a folder by calling
 `FoldersManager.GetCollaborationsAsync(string id, IEnumerable<string> fields = null)` with the ID of the folder.
 
+<!-- sample delete_collaborations_id -->
```
mattwiller: Should be `get_collaborations`

docs/collaborations.md
```diff
@@ -115,6 +120,7 @@ Get Pending Collaborations
 A collection of all the user's pending collaborations can be retrieved with
 `CollaborationsManager.GetPendingCollaborationAsync(IEnumerable<string> fields = null)`.
 
+<!-- sample get_collaborations -->
```
mattwiller: This will conflict with the method above; both hit the same endpoint, just with different query params.  Should this be a variant, e.g. `get_collaborations pending`?

docs/comments.md
```diff
@@ -34,6 +35,7 @@ You can get all of the comments on a file by calling
 `FilesManager.GetCommentsAsync(string id, IEnumerable<string> fields = null)`
 with the ID of the file.
 
+<!-- sample get_comments_id -->
```
mattwiller: Should be `get_comments`

docs/files.md
```diff
@@ -87,6 +90,7 @@ The simplest way to upload a file to a folder is by calling
 `FilesManager.UploadAsync(BoxFileRequest fileRequest, Stream stream, IEnumerable<string> fields = null, TimeSpan? timeout = null, byte[] contentMD5 = null, bool setStreamPositionToZero = true, Uri uploadUri = null)`
 with the upload parameters and a stream of the file contents to upload.
 
+<!-- sample post_files_id_content -->
```
mattwiller: Should be `post_files_content`

docs/metadata-cascade-policies.md
```diff
@@ -47,13 +49,15 @@ Get Metadata Cascade Policies for Folder
 To retrieve a collection of metadata cascade policies within a given folder for the current enterprise, use
 `MetadataCascadePolicyManager.GetAllMetadataCascadePoliciesAsync(string folderId, string ownerEnterpriseId = null, int limit = 100, string nextMarker = null, IEnumerable<string> fields = null, bool autopaginate = false)`
 
+<!-- sample post_metadata_cascade_policies -->
```
mattwiller: Should be `get_metadata_cascade_policies`

docs/metadata-cascade-policies.md
```diff
@@ -47,13 +49,15 @@ Get Metadata Cascade Policies for Folder
...
 ```c#
 BoxCollectionMarkerBased<BoxMetadataCascadePolicy> metadataCascadePolicies = await client.MetadataCascadePolicyManager.GetAllMetadataCascadePoliciesAsync("12345");
 ```
 
 You can also retrieve metadata cascade policies for another enterprise with specific fields to retrieve by using
 `MetadataCascadePolicyManager.GetAllMetadataCascadePoliciesAsync(string folderId, string ownerEnterpriseId, int limit, IEnumberable<string> fields)`
 
+<!-- sample get_metadata_cascade_policies -->
```
mattwiller: This should probably be a variant, e.g. `get_metadata_cascade_policies external_enterprise`

docs/metadata.md
```diff
@@ -423,6 +436,7 @@ Dictionary<string, object> metadata = await client.MetadataManager.
 You can also get all metadata on a folder by calling
 `MetadataManager.GetAllFolderMetadataTemplatesAsync(string folderId)`.
 
+<!-- sample post_folders_id_metadata -->
```
mattwiller: Should be `get_folders_id_metadata`

docs/storage-policies.md
```diff
@@ -63,6 +67,7 @@ To determine which storage policy is assigned to a user, call
 `StoragePoliciesManager.GetAssignmentForTargetAsync(string entityId, string entityType = "user")`
 with the ID of the user.
 
+<!-- sample get_storage_policy_assignments -->
```
mattwiller: Since this "endpoint" requires a certain query param that could someday not be required (to expose true "get all" functionality), should this be a variant?  Maybe something like `get_storage_policy_assignments target`?

docs/storage-policies.md
```diff
@@ -39,6 +41,7 @@ method with the ID of the storage policy to assign and the ID of the user to whi
 > __Note:__ This method will check if an assignment already exists for the user and take appropriate action.
 > It should work regardless of the current status of the user.
 
+<!-- sample post_storage_policy_assignments -->
```
mattwiller: This is the convenience method; you probably want to remove this tag since you already have the endpoint correctly tagged below

docs/storage-policies.md
```diff
@@ -105,6 +112,7 @@ default storage policy for the enterprise, call
 `StoragePoliciesManager.DeleteAssignmentAsync(string assignmentId)` with
 the ID of the assignment to remove.
 
+<!-- sample delete_storage_policy_assignments -->
```
mattwiller: Should be `delete_storage_policy_assignments_id`

docs/storage-policies.md
```diff
@@ -91,6 +97,7 @@ To update a storage policy assignment, for example to update which storage polic
 asisgned to a user, call the `StoragePoliciesManager.UpdateStoragePolicyAssignment(string assignmentId, String policyId)`
 method with the ID of the assignment to update and the new policy ID to assign.
 
+<!-- sample put_storage_policy_assignments -->
```
mattwiller: Should be `put_storage_policy_assignments_id`

docs/tasks.md
```diff
@@ -101,6 +108,7 @@ To assign a task to a user, call
 with the ID of the task to assign and either the ID or login email address of the
 user to whom the task should be assigned.
 
+<!-- sample post_task_assignments_id -->
```
mattwiller: Should be `post_task_assignments`

docs/authentication.md
cbetta: It's the right sample for what I'd like to achieve: show how to init the SDK and get the auth URL.

This maps to the https://account.box.com/api/oauth2/authorize endpoint (`get_authorize`) which isn't really an endpoint but a redirect.

docs/collaborations.md
cbetta: Not really. Turns out the other one was `get_folders_id_collaborations`, and this one doesn't need to be a variation cause I think `get_collaborations` only ever returns pending ones.

docs/metadata-cascade-policies.md
cbetta: Agreed. Removed for now

docs/storage-policies.md
cbetta: Agreed, but for now we don't support variants yet. We can revisit this later when we are going to review every endpoint's documentation in detail.

docs/storage-policies.md
cbetta: Removed

docs/authentication.md
mattwiller: That isn't what this piece of code does, though; it creates a client with a Developer Token and does not generate the authorize URL.  There's a fake redirect URL in the code sample, but it's not being used.  We don't have a sample in this case for what you're looking for.

cbetta: I'm happy to merge this for now. I am going through these endpoints soon in the next few weeks.

PR #563, opened 2019-05-08
Allow download with large byte offsets

### LINKED ISSUE #560: DownloadStreamAsync int offset values
- [x] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
In my application I am using the SDK to download files and if they are above a specified size, I am splitting the requests and downloading in chunks and building the file locally rather than downloading the whole file at once. 
To accomplish this I am setting the `startOffsetInBytes` and `endOffsetInBytes` arguments in the `BoxFilesManager.DownloadStreamAsync` method. 

Unfortunately, those arguments are `int?` types which will not work for any values/file sizes over `int.MaxValue` of  2,147,483,647 bytes or almost 2GB. 

Would it be possible to update that method, or just overload it, using the `long?` type for the `startOffsetInBytes` and `endOffsetInBytes` arguments?

### Versions Used
.NET SDK: 3.16.0
.NET Framework/Core: 4.6.1

mattwiller:
Fixes #560

### DIFF
--- a/Box.V2.Core/Box.V2.Core.csproj
+++ b/Box.V2.Core/Box.V2.Core.csproj
@@ -3,14 +3,14 @@
   <PropertyGroup>
     <Description>Box.V2 for .Net Core</Description>
     <AssemblyTitle>Box.V2.Core</AssemblyTitle>
-    <Version>3.15.0</Version>
+    <Version>3.16.0</Version>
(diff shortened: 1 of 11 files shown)


boxcla:

PR #567, opened 2019-05-17
Throw TimeoutException when request times out

mattwiller:
When the `HttpClient` we use to make API request times out the request, it throws a generic `TaskCanceledException` that makes diagnosing the problem pretty difficult.  To provide users better information about what the exception is meant to represent, the timeout system within the SDK has been refactored to manually cancel the request using a known `CancellationTokenSource` rather than relying on the built-in `HttpClient` timeout.  This allows us to verify whether or not the timeout specifically was the reason for the exception, in a way that should also extend cleanly if/when we allow users to cancel requests with their own cancellation tokens.  It also removes the need for maintaining multiple `HttpClient` instances per timeout value.


boxcla:

Issue #594, opened 2019-09-24
external_app_user_id missing as a BoxUser property


Cpcrook:
- [ x] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
I'd like to be able to set/retrieve the external_app_user_id JSON property off of BoxUser objects.  It's simply missing from the BoxUser object, even though the UsersManager.GetEnterpriseUsers method accepts it as a parameter.

### Versions Used
.NET SDK: 3.20.0

### Steps to Reproduce
There simply isn't a property present.

### Error Message, Including Stack Trace
N/A just make your C# SDK match your REST API.

-----

Again, this is missing and something I'd to actually use within the confines of the C# SDK.  PR incoming, wanted to make sure I had an issue to link to.

PR #595, opened 2019-09-24
Add ExternalAppUserId property to BoxUser model

Cpcrook:
- already part of the BoxUserRequest object, and now it should be possible to pull this back in a user object if it is possible to set it in a request for creation or update.

### DIFF
--- a/Box.V2/Managers/BoxUsersManager.cs
+++ b/Box.V2/Managers/BoxUsersManager.cs
@@ -220,7 +220,7 @@ namespace Box.V2.Managers
         /// Retrieves information about a user in the enterprise. Requires enterprise administration authorization.
         /// </summary>
         /// <param name="userId">The user identifier.</param>
-        /// <returns>Returns the complete user object.</returns>
+        /// <returns>Returns the default representation of the user object.</returns>
         public async Task<BoxUser> GetUserInformationAsync(string userId)
         {
             BoxRequest request = new BoxRequest(_config.UserEndpointUri, userId);
--- a/Box.V2/Managers/BoxUsersManager.cs
+++ b/Box.V2/Managers/BoxUsersManager.cs
@@ -220,46 +220,10 @@ namespace Box.V2.Managers
         /// Retrieves information about a user in the enterprise. Requires enterprise administration authorization.
         /// </summary>
         /// <param name="userId">The user identifier.</param>
-        /// <param name="fields">Attribute(s) to include in the response.</param>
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @Cpcrook, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

sujaygarlanka:
Will take a look at this soon.

Box.V2/Managers/BoxUsersManager.cs
```diff
@@ -220,10 +220,46 @@ public async Task<bool> DeleteEmailAliasAsync(string userId, string emailAliasId
         /// Retrieves information about a user in the enterprise. Requires enterprise administration authorization.
         /// </summary>
         /// <param name="userId">The user identifier.</param>
+        /// <param name="fields">Attribute(s) to include in the response.</param>
         /// <returns>Returns the complete user object.</returns>
-        public async Task<BoxUser> GetUserInformationAsync(string userId)
+        public async Task<BoxUser> GetUserInformationAsync(string userId, IEnumerable<string> fields = null)
         {
-            BoxRequest request = new BoxRequest(_config.UserEndpointUri, userId);
+            //This is supposed to retrieve the "complete" user object according to the API, so it should probably return all of the actual fields available by default.
```
mattwiller: This isn't really the way the SDK is intended to work — it's not guaranteed to produce all possible fields by default.  If additional fields beyond the ones returned from the API by default are required, they should be passed in explicitly via `fields`.  Please remove the section constructing a list of all possible fields below.

Cpcrook: Then I'd suggest simultaneously updating the description of  GetUserInformationAsync as it currently says it "returns the complete user object": https://github.com/box/box-windows-sdk-v2/blob/ca3034ee7663d95c20ce51362d6610f298197d65/Box.V2/Managers/BoxUsersManager.cs#L223

Didn't want to deviate from the existing documentation, hence that inclusion.

Should we call it the "default user object" instead?  Because it's not the mini-user object, but it's also not complete.  I'll be suggesting an edit to the readme.io to make the distinction clearer between different object representations.

mattwiller:
@Cpcrook Thanks again for the contributions!  We're targeting a release next week to push out a new version including your changes.

Cpcrook:
Great, thanks @mattwiller !

Cpcrook:
This closes #594

PR #597, opened 2019-09-26
Add constructor for BoxTrackingCode 

### LINKED ISSUE #593: tracking_codes are not possible to set via the C# SDK
- [ X] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
I want to update or create a user and set the TrackingCodes property (underlying json field is "tracking_codes").
The setters for the BoxTrackingCode object are private, making this impossible, even though it is possible via the REST API directly (once tracking_codes are enabled for your enterprise)

### Versions Used
.NET SDK: 3.20.0

### Steps to Reproduce
Instantiate a BoxTrackingCode.  Try to set any of the properties on it and perform an create/update using it within a BoxUserRequest.

You won't get past compilation to the private setters.

Property or indexer BoxTrackingCode.Name cannot be assigned to it is read-only.

------

I see no good reason this should be read-only when the REST API supports this functionality.  I'll be submitting a PR myself to fix this so I don't have to do reflection to set properties with private setters as a hack.

Cpcrook:
This is intended to allow creation of codes for use in BoxUserRequests, but still leave BoxUser.TrackingCodes immutable.

Closes #593

### DIFF
--- a/Box.V2/Models/BoxTrackingCode.cs
+++ b/Box.V2/Models/BoxTrackingCode.cs
@@ -13,6 +13,19 @@ namespace Box.V2.Models
         public const string FieldValue = "value";
 
         /// <summary>
+        /// Constructor for creating new BoxTrackingCodes with a given name and value, such that they can be created and passed in BoxUserRequests
+        /// </summary>
+        /// <param name="name">Name of a tracking code registered by the enterprise administrator.</param>
+        /// <param name="value">Value of the tracking code.</param>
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @Cpcrook has signed the CLA. Thanks for the pull request!

mattwiller:
@Cpcrook Thanks for the PR! 🎉

Issue #601, opened 2019-10-11
Improvement: optional fields param on UserManager.GetUserInformationAsync


Cpcrook:
This would be consistent with the underlying REST API (including the collection of Postman examples).

Currently not supported by the Windows .NET SDK V2, despite the docs [here](https://developer.box.com/docs/sdks) saying the .NET SDK is in parity (presumably) with the underlying REST API.

nidoran:
https://github.com/box/box-windows-sdk-v2/pull/605

PR #604, opened 2019-10-11
Add optional fields parameter for GetUserInformationAsync call

Cpcrook:
…for consistency with underlying REST API

Pretty straight forward and should handle issue #601


boxcla:
Verified that @Cpcrook has signed the CLA. Thanks for the pull request!

carycheng:
@Cpcrook, thank you for the pr! Since it is a breaking change to change the interface, we would probably have to have an override for GetUserInformationAsync() with the fields. This is something our team can definitely take over and bring to the end.

carycheng:
#605

carycheng:
#605

PR #631, opened 2020-02-12
Fix Authentication Request Retries

PJSimon:


boxcla:
Verified that @PJSimon has signed the CLA. Thanks for the pull request!

Box.V2/JWTAuth/BoxJWTAuth.cs
```diff
@@ -145,27 +149,61 @@ public string UserToken(string userId)
...
+                        ||
+                        (ex.StatusCode == HttpStatusCode.Accepted && retryAfterHeader != null))
+                        && retryCounter++ < HttpRequestHandler.RetryLimit)
+                    {
+                        // Before we retry the JWT Authentication request, we must regenerate the JTI claim with an updated datetime.
+                        var serverDate = ex.ResponseHeaders != null ? ex.ResponseHeaders.Date : null;
+                        if (serverDate.HasValue)
+                        {
+                            var date = serverDate.Value;
+                            assertion = ConstructJWTAssertion(subId, subType, date.LocalDateTime);
```
sujaygarlanka: Do we want to add the delay to the expiration time?

Box.V2/JWTAuth/BoxJWTAuth.cs
```diff
@@ -145,27 +149,61 @@ public string UserToken(string userId)
...
+                        }
+                        else
+                        {
+                            assertion = ConstructJWTAssertion(subId, subType, DateTime.Now);
+                        }
+
+                        TimeSpan delay = expBackoff.GetRetryTimeout(retryCounter);
```
sujaygarlanka: I think the retry-after header time should be used here for the delay if available.

Box.V2/JWTAuth/BoxJWTAuth.cs
```diff
@@ -145,27 +149,61 @@ public string UserToken(string userId)
...
+                    if ((ex.StatusCode == HttpRequestHandler.TooManyRequests
+                        ||
+                        ex.StatusCode == HttpStatusCode.InternalServerError
+                        ||
+                        ex.StatusCode == HttpStatusCode.BadGateway
+                        ||
+                        ex.StatusCode == HttpStatusCode.ServiceUnavailable
+                        ||
+                        ex.StatusCode == HttpStatusCode.GatewayTimeout
+                        ||
```
sujaygarlanka: May want to retry on the 400 error where the Unix time on the user's machine does not match the server.

Box.V2/JWTAuth/BoxJWTAuth.cs
PJSimon: Done!

Box.V2/JWTAuth/BoxJWTAuth.cs
PJSimon: Done!

Box.V2/JWTAuth/BoxJWTAuth.cs
PJSimon: Done!

PR #633, opened 2020-02-13
Add metadata query

sujaygarlanka:


boxcla:
Verified that @sujaygarlanka has signed the CLA. Thanks for the pull request!

Box.V2/Managers/BoxResourceManager.cs
```diff
@@ -225,6 +227,37 @@ protected async Task<BoxCollectionMarkerBasedV2<T>> AutoPaginateMarkerV2<T>(BoxR
...
+            do
+            {
+                var response = await ToResponseAsync<BoxCollectionMarkerBased<T>>(request).ConfigureAwait(false);
+                var newItems = response.ResponseObject;
+                allItemsCollection.Entries.AddRange(newItems.Entries);
+                allItemsCollection.Order = newItems.Order;
+
+                dynamic body = JObject.Parse(request.Payload);
+                body.marker = newItems.NextMarker;
+                request.Payload = body.ToString();
```
PJSimon: Except for these three lines, this function is almost the same as `AutoPaginateMarker<T>`.  Can you EASILY Refactor/Extract these three lines into a new method and consolidate the common code between the two functions?  If non-trival, then let's just keep it as-is.

Box.V2.Test/BoxMetadataManagerTest.cs
```diff
@@ -562,5 +563,66 @@ public async Task SetFolderMetadataAsync_Create_Error()
...
...eated_at\":\"2017-01-06T17:59:01-08:00\",\"content_modified_at\":\"2017-01-06T17:59:01-08:00\",\"created_by\":{\"type\":\"user\",\"id\":\"193973366\",\"name\":\"Box Admin\",\"login\":\"admin@company.com\"},\"modified_by\":{\"type\":\"user\",\"id\":\"193973366\",\"name\":\"Box Admin\",\"login\":\"admin@company.com\"},\"owned_by\":{\"type\":\"user\",\"id\":\"193973366\",\"name\":\"Box Admin\",\"login\":\"admin@company.com\"},\"shared_link\":null,\"parent\":{\"type\":\"folder\",\"id\":\"16125613433\",\"sequence_id\":\"0\",\"etag\":\"0\",\"name\":\"2017\"},\"item_status\":\"active\"},\"metadata\":{\"enterprise_123456\":{\"someTemplate\":{\"$parent\":\"file_161753469109\",\"$version\":0,\"customerName\":\"Phoenix Corp\",\"$type\":\"someTemplate-3d5fcaca-f496-4bb6-9046-d25c37bc5594\",\"$typeVersion\":0,\"$id\":\"ba52e2cc-371d-4659-8d53-50f1ac642e35\",\"amount\":100,\"claimDate\":\"2016-04-10T00:00:00Z\",\"region\":\"West\",\"$typeScope\":\"enterprise_123456\"}}}}],\"next_marker\":\"AAAAAmVYB1FWec8GH6yWu2nwmanfMh07IyYInaa7DZDYjgO1H4KoLW29vPlLY173OKsci6h6xGh61gG73gnaxoS+o0BbI1/h6le6cikjlupVhASwJ2Cj0tOD9wlnrUMHHw3/ISf+uuACzrOMhN6d5fYrbidPzS6MdhJOejuYlvsg4tcBYzjauP3+VU51p77HFAIuObnJT0ff\"}";
```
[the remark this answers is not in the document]
sujaygarlanka: Got them from the metadata query doc. Other data in the response is seems made up, so I believe this data is as well.

Box.V2/Managers/BoxResourceManager.cs
sujaygarlanka: I initially tried to put it into a single method, but `AutoPaginateMarker<T>` expects `T` to be a `BoxEntity` and the metadata endpoint required creating a new model that was not a `BoxEntity`. So I created a new method for this case.

Issue #638, opened 2020-03-03
a deadlock occurs when obtaining an access token.


sin5:
### Description of the Issue
In my Windows Forms application, a deadlock occurs when obtaining an access token.
I think that "ConfigureAwait (false)" is necessary for "await getResponse (request, isStream, httpRequest)" on line 40 of Box.V2/Request/HttpRequestHandler.cs

### Versions Used
.NET SDK:4.7.1

### Steps to Reproduce
```
        private void button1_Click(object sender, EventArgs e)
        {
            StreamReader reader = new StreamReader("./config.json");
            string json = reader.ReadToEnd();
            IBoxConfig config = BoxConfig.CreateFromJsonString(json);
            BoxJWTAuth boxJWT = new BoxJWTAuth(config);
            string token = boxJWT.AdminToken();  // <<== deadlock
        }
```

### Error Message, Including Stack Trace
No message due to deadlock.
The getResponse method of HttpRequestHandler class is trying to return "StatusCode: 200, ReasonPhrase: 'OK', Version: 1.1, Content: System.Net.Http.StreamContent, Headers:".

PJSimon:
@sin5, thank you for opening this issue.  We will take a look ASAP and get back to you soon.  Thanks!

sujaygarlanka:
Hello @sin5, Thank you for investigating this issue. I took a look at line 40 in `HttpRequestHandler.cs` and I believe you are correct. We need to either add an `ConfigureAwait(false)` there or add an `await` on this line https://github.com/box/box-windows-sdk-v2/blob/86ded84643df0375b316d30e6af5e379622a3faf/Box.V2/JWTAuth/BoxJWTAuth.cs#L276. Could you please attach your application or a sample of the code that was failing for you? It will make it easier for us to test when we implement the fix.

tjones0808:
once i hit the line of getting a token my app stops responding. has this been fixed on nuget yet?

sin5:
Hello @sujaygarlanka , 　Thank you for your reply.
Attach a sample solution that will deadlock.

In my environment, I added await to the indicated location in BoxJWTAuth.cs, but it seems deadlock has occurred.

[BoxTest_638.zip](https://github.com/box/box-windows-sdk-v2/files/4341724/BoxTest_638.zip)

tjones0808:
> Hello @sujaygarlanka , 　Thank you for your reply.
> Attach a sample solution that will deadlock.
> 
[...]

PR #646, opened 2020-04-08
Fix deadlock issue for JWT Auth in UI elements

sujaygarlanka:
When a user calls `BoxJWT.AdminToken()` to get a token in a UI element, a deadlock issue will occur because the underlying async request methods in `HttpRequestHandler.cs` don't have `Configure.await(false)` appended to them. This PR is in regards to this issue (#638).

Why this is a fix is detailed in this [article](https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f).

### DIFF
--- a/Box.V2/Request/HttpRequestHandler.cs
+++ b/Box.V2/Request/HttpRequestHandler.cs
@@ -38,8 +38,8 @@ namespace Box.V2.Request
 
                 HttpRequestMessage httpRequest = getHttpRequest(request, isMultiPartRequest, isBinaryRequest);
                 Debug.WriteLine(string.Format("RequestUri: {0}", httpRequest.RequestUri));
-                HttpResponseMessage response = await getResponse(request, isStream, httpRequest);
-                BoxResponse<T> boxResponse = await getBoxResponse<T>(isStream, response);
+                HttpResponseMessage response = await getResponse(request, isStream, httpRequest).ConfigureAwait(false);
+                BoxResponse<T> boxResponse = await getBoxResponse<T>(isStream, response).ConfigureAwait(false);
 
                 return boxResponse;
             }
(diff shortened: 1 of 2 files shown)


sam2332:
does this fix the hang in asp.net
for 
```
            var sdk = new BoxJWTAuth(config);
            var token = sdk.AdminToken();
```

sam2332:
`
Models\Request\BoxUserRequest.cs(129,16): error CS0246: The type or namespace name 'BoxNotificationEmailField' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\box-windows-sdk-v2\Box.V2\Box.V2.csproj]
`

I was having an issue with this also, did something get committed to the master branch that was not working?

sam2332:
https://github.com/box/box-windows-sdk-v2/commit/e2c679db7a21431b379dfb39cd8a590f28d95e3a

sujaygarlanka:
@sam2332 This should fix your hang issue. We are looking into the Notification Field issue.

BryceBarbara:
This isn't actually the proper solution. Yes, it _technically_ does fix the deadlock but now we're still making async calls from non-async methods and retrieving the result.

Check out [this article's](https://markheath.net/post/async-antipatterns) 4th section titled "Blocking on tasks with .Result or .Wait".

sujaygarlanka:
@BryceBarbara Thank you for the feedback. I read the article. Correct me if I am wrong, but you are suggesting we make `JWTAuthPost` and subsequent methods async to avoid the need to add a `ConfigureAwait(false)`. If so, I think this makes sense. However, there are two issues with doing so currently. 

- Our final method `AdminToken()` would have to be async. This would be a breaking change for our SDK. It could cause some developers' code to stop working. 
- Also, even if we were to make it async, it makes sense to add the `ConfigureAwait(false)` to allow developers to call the `AdminToken()` method synchronously, which may be necessary in some scenarios. 

I hope I understood and answered your question thoroughly. If I didn't, feel free to comment.

BryceBarbara:
I know I'm like super late to this, sorry about that. Work's been crazy!

Anyways, the proper solution would be to have an async version and non-async version. With other SDK's that have needed to retroactively add async versions of functions the convention was to simply add "Async" to the name. So `AdminTokenAsync` would exist alongside `AdminToken`.

kimdimasan:
I agree, we can just have an awaitable async method sdk.AdminClientAsync(token).

Issue #669, opened 2020-05-21
Regression Bug: JsonSerializationException for GetCurrentUserInformationAsync


DaleyKD:
- [X] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [X] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [X] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [X] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue
Just upgrade from Box.V2.Core 3.**22**.0 to 3.**23**.0. A regression issue has bubbled up.

I can no longer call `var user = await boxClient.UsersManager.GetCurrentUserInformationAsync();` without the a Newtonsoft deserialization issue.  See Stack Trace below for further details.

### Steps to Reproduce
```csharp
internal BoxClient GetBoxUserClient(string userID)
{
    var boxConfig = new BoxConfig(this.ClientID, this.ClientSecret, this.EnterpriseID, this.PrivateKey, this.PrivateKeyPassphrase, this.PublicKeyID);
    var boxJwt = new BoxJWTAuth(boxConfig);
    var userToken = boxJwt.UserToken(userID);
    return boxJwt.UserClient(userToken, userID);
}
```

```csharp
var boxClient = GetBoxUserClient("validuser@mydomain.com");
var user = await boxClient.UsersManager.GetCurrentUserInformationAsync();
```

Use the above code in version 3.22.0: Works.
Use the above code in version 3.23.0: Exception thrown.

### Expected Behavior
Not have an exception thrown.

### Error Message, Including Stack Trace
```
Newtonsoft.Json.JsonSerializationException: Cannot deserialize the current JSON array (e.g. [1,2,3]) into type 'Box.V2.Models.BoxNotificationEmail' because the type requires a JSON object (e.g. {"name":"value"}) to deserialize correctly.
To fix this error either change the JSON to a JSON object (e.g. {"name":"value"}) or change the deserialized type to an array or a type that implements a collection interface (e.g. ICollection, IList) like List<T> that can be deserialized from a JSON array. JsonArrayAttribute can also be added to the type to force it to deserialize from a JSON array.
Path 'notification_email', line 1, position 451.
[...]

PR #677, opened 2020-06-26
Fix bug with notification email field deserializing for `BoxUser`

sujaygarlanka:
The `notification_email` field for the user object is an object when a value exists and is an empty array when it isn't. This was causing an issue when deserializing to a `BoxUser`. The code implemented converts the empty array to a null before deserializing. This is also forwards compatible. So if the array is changed to a null when the API is fixed, this will continue to work. This PR addresses issue #669.

### DIFF
--- a/Box.V2.Test/BoxUsersManagerTest.cs
+++ b/Box.V2.Test/BoxUsersManagerTest.cs
@@ -133,7 +133,8 @@ namespace Box.V2.Test
                 ""my_tags"": [
                     ""important""
                 ],
-                ""hostname"": ""https://example.app.box.com/""
+                ""hostname"": ""https://example.app.box.com/"",
+                ""notification_email"": []
             }";
             Handler.Setup(h => h.ExecuteAsync<BoxUser>(It.IsAny<IBoxRequest>()))
                 .Returns(Task.FromResult<IBoxResponse<BoxUser>>(new BoxResponse<BoxUser>()
@@ -143,7 +144,7 @@ namespace Box.V2.Test
                 }));
 
             /*** Act ***/
-            string[] fields = { "name", "timezone", "is_external_collab_restricted", "my_tags", "hostname" };
+            string[] fields = { "name", "timezone", "is_external_collab_restricted", "my_tags", "hostname", "notification_email" };
             BoxUser user = await _usersManager.GetCurrentUserInformationAsync(fields);
 
             /*** Assert ***/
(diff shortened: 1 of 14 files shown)


PJSimon:
@sgarlanka could you please revert the changes on:
Box.V2.Test.Integration/BoxFoldersManagerTestIntegration.cs
and
Box.V2.Samples.Core.AppUser.Create/Program.cs

Box.V2/Converter/BoxItemConverter.cs
```diff
@@ -166,6 +166,11 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
 
             // Load JObject from stream
             JObject jObject = JObject.Load(reader);
+            // The notification_email field for the user object is an object when a value exists and is an empty array when it isn't. The code below converts the array to null to avoid deserialization issues.
```
PJSimon, line 169:
```suggestion
            // The notification_email field for the user object is an object when a value exists and is an empty array when it isn't. The code below converts the empty array to a null value to avoid deserialization issues.  
```

Box.V2/Converter/BoxItemConverter.cs
```diff
@@ -166,6 +166,11 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
 
             // Load JObject from stream
             JObject jObject = JObject.Load(reader);
+            // The notification_email field for the user object is an object when a value exists and is an empty array when it isn't. The code below converts the array to null to avoid deserialization issues.  
+            if (jObject["type"] != null && jObject["type"].ToString() == "user" && jObject["notification_email"] != null && jObject["notification_email"].Type == JTokenType.Array)
```
PJSimon, line 170: Should we check that the array is also empty, just to be sure?

Box.V2.Test/BoxUsersManagerTest.cs
```diff
@@ -51,7 +51,7 @@ public async Task GetUserInformation_WithField_ValidResponse_ValidUser()
         {
             /*** Arrange ***/
             IBoxRequest boxRequest = null;
-            string responseString = "{\"type\": \"user\", \"id\": \"12345\", \"status\": \"active\"}";
+            string responseString = "{\"type\": \"user\", \"id\": \"12345\", \"status\": \"active\", \"notification_field\": []}";
```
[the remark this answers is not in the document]
PJSimon: Also, can we parameterize the responseString for this method so we can call it twice: once with this "empty array" string and once with "null" for notification_email?

Box.V2/Converter/BoxItemConverter.cs
sujaygarlanka: Whether it is empty or not, we should make it null or it will throw an error.

Box.V2.Test/BoxUsersManagerTest.cs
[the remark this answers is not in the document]
sujaygarlanka: I test the null case in a separate test method. It would make sense to parameterize if this was long term, but we are probably going to remove this in a quarter or two when the API is fixed.

Issue #636, opened 2020-02-21
BoxUserManager.MoverUserFolderAsync configurable timeout


mkiosugadaikichi:
When user owns many files,
BoxUserManager.MoverUserFolderAsync method takes longer time.

Current implementation, timeout is fixed default value.
HttpClient default timeout value is 100 seconds.
I think configurable timeout is useful.

PJSimon:
@mkiosugadaikichi, thank you for opening this issue.  We will take a look ASAP and get back to you soon.  Thanks!

sujaygarlanka:
@mkiosugadaikichi This issue has been solved by #681.

PR #681, opened 2020-07-20
Add ability to set a request timeout when updating folder information and moving a user's folder

sujaygarlanka:
Add ability to set a request timeout for `FoldersManager.UpdateInformationAsync` and `UsersManager.MoverUserFolderAsync`


sujaygarlanka:
#636

PR #717, opened 2021-01-07
Add filter fields to get file version retentions

sujaygarlanka:
Add filter params to `BoxRetentionPoliciesManager.GetFileVersionRetentionsAsync()`

### DIFF
--- a/Box.V2.Test/BoxRetentionPoliciesManagerTest.cs
+++ b/Box.V2.Test/BoxRetentionPoliciesManagerTest.cs
@@ -1,13 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
 using Box.V2.Managers;
 using Box.V2.Models;
-using Box.V2.Config;
-using Box.V2.Auth;
+using Box.V2.Models.Request;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
-using System.Threading.Tasks;
-using System.Collections.Generic;
-using Box.V2.Models.Request;
-using System;
 
 namespace Box.V2.Test
 {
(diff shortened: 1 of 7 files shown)


Box.V2/Models/BoxRetentionPolicy.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Newtonsoft.Json;
+using Newtonsoft.Json;
```
PJSimon, line 1:
```suggestion
```

PJSimon: Looks like this is no longer needed.

Box.V2/Models/BoxEnums.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
PJSimon, line 1:
```suggestion
```

PJSimon: This doesn't appear to be needed and can be deleted.  Please confirm.

Box.V2.Test/BoxRetentionPoliciesManagerTest.cs
```diff
@@ -7,6 +7,7 @@
 using System.Threading.Tasks;
 using System.Collections.Generic;
 using Box.V2.Models.Request;
+using System;
```
PJSimon, line 10: Can you sort the using statement (I think you can right click in VS and have them sorted).
This needs to move to line 7, above other `System.` using statements.
Line 9 Needs to move up, too...

Box.V2/Models/BoxRetentionPolicy.cs
PJSimon: Sorry, this should have said, "This looks like a Byte Order Mark update on a file with no other changes.  Can you please reset this file to remove it from the PR?"

Box.V2.Test/BoxRetentionPoliciesManagerTest.cs
```diff
@@ -1,12 +1,11 @@
-﻿using Box.V2.Managers;
+using System;
```
PJSimon, line 1: Sort using statements in Visual Studio to put them in alphabetical order.

PJSimon (approved):

PR #721, opened 2021-02-16
Add support for search param to get shared link items

sujaygarlanka:


Box.V2/Models/BoxSearchResult.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>
+        /// The type of the item
+        /// </summary>
+        [JsonProperty(PropertyName = "type")]
+        public string Type { get; private set; }
+
+        /// <summary>
+        /// The optional shared link through which the user has access to this item.
+        /// This value is only returned for items for which the user has recently accessed the file through a shared link. For all other items this value will return nil.
```
PJSimon, line 16: fsdsf
```suggestion
        /// This value is only returned for items for which the user has recently accessed the file through a shared link. For all other items this value will return null.
```

PJSimon (approved):

PR #739, opened 2021-08-03
Copy stream for upload requests and remove expect header

mwwoda:
Stream is being closed (disposed) when request is sent by HttpClient. See https://github.com/dotnet/runtime/issues/14612

> When a request completes, HttpClient disposes the request content so the user doesn't have to. This also ensures that a HttpContent object is only sent once using HttpClient (similar to HttpRequestMessages that can also be sent only once)

It was actually resolved in .NET Core (https://github.com/dotnet/corefx/pull/19082), but since we need to support the .NET Framework too, another fix is required.

The drawback of copying stream is that we allocate additional memory for the copied stream (basically stream size is doubled).

Another fix that could be applied instead would be to check if token is valid before making the upload request.

### DIFF
--- a/Box.V2/Request/HttpRequestHandler.cs
+++ b/Box.V2/Request/HttpRequestHandler.cs
@@ -69,42 +69,40 @@ namespace Box.V2.Request
                 while (true)
                 {
                     using (HttpRequestMessage httpRequest = getHttpRequest(request, isMultiPartRequest, isBinaryRequest))
+                    using (HttpResponseMessage response = await getResponse(request, isStream, httpRequest).ConfigureAwait(false))
                     {
                         Debug.WriteLine(string.Format("RequestUri: {0}", httpRequest.RequestUri));
-                        using (HttpResponseMessage response = await getResponse(request, isStream, httpRequest).ConfigureAwait(false))
+                        //need to wait for Retry-After seconds and then retry request
+                        var retryAfterHeader = response.Headers.RetryAfter;
+
+                        // If we get a retryable/transient error code and this is not a multi part request (meaning a file upload, which cannot be retried
+                        // because the stream cannot be reset) and we haven't exceeded the number of allowed retries, then retry the request.
+                        // If we get a 202 code and has a retry-after header, we will retry after
+                        if (!isMultiPartRequest &&
+                            (response.StatusCode == TooManyRequests
+                            ||
+                            response.StatusCode == HttpStatusCode.InternalServerError
+                            ||
+                            response.StatusCode == HttpStatusCode.BadGateway
+                            ||
+                            response.StatusCode == HttpStatusCode.ServiceUnavailable
+                            ||
+                            response.StatusCode == HttpStatusCode.GatewayTimeout
+                            ||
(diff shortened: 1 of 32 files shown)


mhagmajer (dismissed):

Box.V2/Request/ReusableContent.cs
```diff
@@ -0,0 +1,35 @@
...
+        private HttpContent _innerContent;
+
+        public ReusableContent(Stream stream)
+        {
+            _innerContent = new StreamContent(stream);
+        }
+
+        protected override async Task SerializeToStreamAsync(Stream stream, TransportContext context)
+        {
+            await _innerContent.CopyToAsync(stream);
```
PJSimon, line 19: I do feel like it's not ideal to do this copy operation.  I don't fully understand it, TBH.  When exactly is this method called? When does the `stream` property get disposed?  Can we dispose `stream` at the end of this method to free up memory sooner and explicitly?

Maybe the token method you mentioned would be "safer."  Where would that check need to be inserted?

Good job researching this and coming up with an otherwise elegant solution!

Box.V2/Request/HttpRequestHandler.cs
```diff
@@ -66,41 +68,44 @@ public async Task<IBoxResponse<T>> ExecuteAsync<T>(IBoxRequest request)
...
-                        await Task.Delay(delay);
-                    }
-                    else
+                    using (HttpRequestMessage httpRequest = getHttpRequest(request, isMultiPartRequest, isBinaryRequest))
                     {
-                        BoxResponse<T> boxResponse = await getBoxResponse<T>(isStream, response).ConfigureAwait(false);
-
-                        return boxResponse;
+                        Debug.WriteLine(string.Format("RequestUri: {0}", httpRequest.RequestUri));
+                        using (HttpResponseMessage response = await getResponse(request, isStream, httpRequest).ConfigureAwait(false))
```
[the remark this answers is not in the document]
mwwoda: Good point!

Box.V2/Request/ReusableContent.cs
mwwoda: `SerializeToStreamAsync` is called internally by the `HttpClient` during `sendAsync(..)` method. We cannot dispose `_innerContent` yet as it will be used during the retry in case the request fails (`stream` should be disposed by `httpClient`). If we don't copy this, the `_innerContent` will be disposed internally and retry will fail. This is the behaviour of the `HttpClient` (see links in description) that dispose the resources it's not the owner of.

The token method I mentioned can solve stream issue in some cases and save memory by preventing copying, but still they are some cases that the retry would fail. By calling token just before the upload request, we can ensure that it will be valid. 401 code should be no longer present. There is still a possibility that the request will fail for different reason (e.g. 500 code from server). In this case the Stream will be disposed and retry will fail every time.  

The solution is not perfect, but I don't see any alternative if we want to stick to the current `HttpClient` and solve the retry problem.

mhagmajer (approved):

Box.V2/Request/ReusableContent.cs
PJSimon: Makes sense!  Thanks!

PJSimon (approved):

PR #740, opened 2021-08-04
Add new method for `Create Terms of Service Status`

mwwoda:


mhagmajer (approved):
Let's also mark `BoxTermsOfServiceUserStatusesRequest` class as obsolete

mwwoda:
`BoxTermsOfServiceUserStatusesRequest` is not exactly obsolete as it also used in another api call (`UpdateTermsofServiceUserStatusesAsync(string tosId, bool isAccepted)`) that is not obsolete.

PR #660, opened 2020-04-29
Switch over to DateTimeOffset

BryceBarbara:
This is to address a long time shortcoming of the SDK which is the improper handling of dates and times.

This should address #394

PR #749, opened 2021-09-14
Use DateTimeOffset instead of DateTime

### LINKED ISSUE #394: Use `DateTimeOffset` instead of `DateTime`
I quickly went over the current usages of `DateTime` and it seems that all of them are there to represent an exact, specific moment in time (the file was created at time X, the folder was shared at time Y, the session expires at time Z etc). 

Per the [docs](https://docs.microsoft.com/en-us/dotnet/standard/datetime/choosing-between-datetime), `DateTimeOffset` should be used in such cases (emphasis mine):

> The DateTimeOffset type includes all of the functionality of the DateTime type along with time zone awareness. This makes it suitable for applications that do the following:
> **Uniquely and unambiguously identify a single point in time**. The DateTimeOffset type can be used to unambiguously define the meaning of "now", to log transaction times, to log the times of system or application events, and to record file creation and modification times.

Regarding when to use `DateTime`, here's a good excerpt from a related [StackOverflow answer](https://stackoverflow.com/questions/4331189/datetime-vs-datetimeoffset):

> I see DateTime as being useful when you want to deal with dates only, times only, or deal with either in a generic sense. For example, if you have an alarm that you want to go off every day at 7 am, you could store that in a DateTime utilizing a DateTimeKind of Unspecified because you want it to go off at 7am regardless of DST. But if you want to represent the history of alarm occurrences, you would use DateTimeOffset.

mwwoda:
Resolves #394. Continue of #660.


mhagmajer (approved):
:shipit:

PR #753, opened 2021-09-30
Box exceptions rework according to the reference architecture

mwwoda:
A distinction has been made between errors for coding errors (maybe we can think of a better name(?)), API errors and common errors based on reference architecture. This should make it easier to handle e.g. retries. Also got rid of unnecessary setters. 

Also solves SDK-1528. The Error object should always be present in case of an API error (otherwise it's a problem in the API), but just in case, I added a default value for the code and message where necessary.

### DIFF
--- a/Box.V2.Core/Box.V2.Core.csproj
+++ b/Box.V2.Core/Box.V2.Core.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Box.V2 for .Net Core</Description>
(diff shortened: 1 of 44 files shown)


antusus (approved):
I'm not sure about the name. We are mostly using this exception when we are validating user input. So maybe BoxValidationException or BoxInvalidArgumentException, or BoxRuntimeException...

antusus (approved):

mwwoda:
> I'm not sure about the name. We are mostly using this exception when we are validating user input. So maybe BoxValidationException or BoxInvalidArgumentException, or BoxRuntimeException...

This is true for the input, but we also throw this exception after the API cal. According to our reference architecture, we should use these exceptions when encoding or decoding an http request. We could also distinguish between Coding and Validation errors, but I'm not sure if there are enough benefits to do this right now.

PR #763, opened 2021-10-18
style: add code style standards to .editorconfig

### LINKED ISSUE #745: Coding standards in box-windows-sdk
When wanting to work on the Box Windows SDK, it would be nice to have a standard (e.g. editorconfig or stylecop) to solidify coding standards. Currently, the repo is inconsistent about simple things, such as when to use var or the object type. This makes it a bit problematic for new devs to know what should be var and what shouldn't be. Especially since lack of standards could lead to the PR failing a review

A clear standard codified in an editor config file or style cop

A published standard in the readme

mwwoda:
Fixes #745. Code formatting was done using https://github.com/dotnet/format.

### DIFF
--- a/Box.V2.Test.Integration/BoxCollaborationsManagerTestIntegration.cs
+++ b/Box.V2.Test.Integration/BoxCollaborationsManagerTestIntegration.cs
@@ -104,7 +104,6 @@ namespace Box.V2.Test.Integration
         }
 
         [TestMethod]
-        [System.Obsolete]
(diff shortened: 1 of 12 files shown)


mwwoda:
> In most cases when rule is violated we produce warning. In Java we are failing build. Maybe it make sense to change those rules to errors and fix the styling?

If we change this to errors, they will be displayed in VisualStudio as a compiler error, which is confusing for developers. The usual practice is to use warnings/suggestions and then go with low/no warning policy in CI.

PR #783, opened 2021-12-13
chore: release script changes

mwwoda:
Fixed several issues found during testing with the release VM. Also changed the approach to creating the release draft first and attaching artifacts to it as soon as they are created to avoid storing artifacts between steps.


antusus (approved):
OK

PR #785, opened 2021-12-14
chore: fix release script

mwwoda:


build/bump_version.ps1
```diff
@@ -100,9 +100,11 @@ if($DryRun){
     git remote set-url origin $RepoLink
     git config user.name $GithubUsername
     git config user.email $GithubEmail
+    $lastBranch = git rev-parse --abbrev-ref HEAD
     git branch -D $NEXT_VERSION_TAG
     git checkout -b $NEXT_VERSION_TAG
-    git commit -am $NEXT_VERSION_TAG
+    $commitTitle = "chore: release " + $NEXT_VERSION_TAG
```
antusus, line 106: I think it should be `chore(release): tag`

antusus (commented):
I'm not sure about release commit name. Otherwise it looks good.

build/bump_version.ps1
mwwoda: Have we agreed on some standard for release commit names? I don't usually use scopes when following conventional-commits in the sdks. I could change it, but I don't see much difference.

antusus (approved):
👍

PR #786, opened 2021-12-16
test: add new integration test project

mwwoda:
The project name is temporary. We will change it after we get rid of the old one.


antusus (approved):
👍

Issue #774, opened 2021-11-19
Client credentials support


tallesl:
Is there support for the `Server Authentication (Client Credentials Grant)` authentication method?

![image](https://user-images.githubusercontent.com/3655047/142629703-6ace8ebf-ce94-4b50-8333-90a5c5e12e81.png)

arjankowski:
Hi @tallesl

Currently our SDKs doesn't support client credentials, but we have a plan to do this in the near future.

@arjankowski

mwwoda:
Resolved via #799.

PR #799, opened 2022-02-03
feat: add Client Credentials Grant auth support

### LINKED ISSUE #719: Add support for new Client Credentials Grant Type
Per blog post (https://support.box.com/hc/en-us/articles/360056830074-Authenticate-apps-with-Client-Credentials-Grant-Type) the API now supports a Client Credential grant type that simplifies some app access. The post is from Nov 2020 and the last commit on the repo was Oct 2020.

Documentation and method support to allow a token to be generated with simply the client_id, client_secret, and enterprise_id (or user as box_subject_type).

Manually generating an access token via RestSharp and a custom call. This requires extra objects and having to parse in the returned access token to the OAuthSession object. No support for autorenew or generating a new token without calling back to custom RestSharp method.

mwwoda:
Resolves #719 and #774.

### DIFF
--- a/Box.V2/Config/Constants.cs
+++ b/Box.V2/Config/Constants.cs
@@ -160,6 +160,8 @@ namespace Box.V2.Config
         public const string AuthHeaderKey = "Authorization";
         public const string V1AuthString = "BoxAuth api_key={0}&auth_token={1}";
         public const string V2AuthString = "Bearer {0}";
+        public const string BearerTokenType = "bearer";
+        public const int AccessTokenExpirationTime = 3600; // seconds
 
         /*** Return types ***/
         public const string TypeFile = "file";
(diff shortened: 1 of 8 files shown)


Box.V2/Config/Constants.cs
```diff
@@ -160,6 +160,8 @@ public static class Constants
         public const string AuthHeaderKey = "Authorization";
         public const string V1AuthString = "BoxAuth api_key={0}&auth_token={1}";
         public const string V2AuthString = "Bearer {0}";
+        public const string BearerTokenType = "bearer";
+        public const int AccessTokenExpirationTime = 3600; // seconds
```
antusus, line 164: Maybe this should be called `ClientCredentialsAccessTokenExpirationTime` ?

antusus (approved):
👍

Box.V2/Config/Constants.cs
mwwoda: The expiration time value is tied directly to the OAuth, not to the CCG itself - JWT Auth also has the same expiration time.

PR #806, opened 2022-02-17
test: migrate old tests to the new test framework

### LINKED ISSUE #399: Integration Tests
1. There is no guidance for integration test configuration, one has to run, fail, and then go investigate what's missing
1. `ClientID` and `ClientSecret` are constants in `BoxResourceManagerTestIntegration`, which might lead to accidental check-in of secrets. The easiest solution would be to read them from app.config and then each developer could define it in his `machine.config` and never check it in. An environment variable would be another option.
1. Getting the full configuration from `JSON_CONFIG` is not a good idea due to environment variable length limits. One solution would be to point it to a file.
1. Even after I fixed all the above, most tests are failing. Here are the full test results: [ohads_OHADS-DESKTOP 2017-09-25 17_11_13.zip](https://github.com/box/box-windows-sdk-v2/files/1329762/ohads_OHADS-DESKTOP.2017-09-25.17_11_13.zip)

mwwoda:
Migrate all tests that were previously running on AppVeyor. Fixes #399.


antusus (commented):
I think that maybe we would like and adopt more BDD style testing?
When removing elements we should check that method is calling API, I think we might get deleted object and verify it was deleted?

antusus (approved):
👍

PR #809, opened 2022-02-24
chore: reorder changelog sections during release

mwwoda:


build/bump_version.ps1
```diff
@@ -81,6 +81,60 @@ $NEXT_VERSION_TAG = "v" + "$NEXT_VERSION"
...
+        $sections += $matches[0]
+    }
+}
+
+$orderedSections = new-object string[] $sections.Length
+$currentSection = $null
+$previousSectionIndex = 0
+
+foreach($line in Get-Content $CHANGELOG_PATH) {
+    if($line -match "## [[0-9]+\.[0-9]+\.[0-9]+]"){
```
arjankowski, line 102: It should be "#{2,3} [[0-9]+\.[0-9]+\.[0-9]+]" because if there is only release with fixes then # character will occur 3 times

mwwoda: Good catch!

arjankowski (approved):
👍

arjankowski (approved):

PR #815, opened 2022-04-05
fix: simplify base `urls` usage

mwwoda:


Box.V2/Config/BoxConfigBuilder.cs
```diff
@@ -133,6 +134,7 @@ public BoxConfigBuilder SetBoxApiHostUri(Uri boxApiHostUri)
         /// </summary>
         /// <param name="boxAccountApiHostUri">BoxAPI account host uri.</param>
         /// <returns>this BoxConfigBuilder object for chaining</returns>
+        [Obsolete("Use BoxAuthAuthorizeApiUri instead")]
```
antusus, line 137: Shouldn't this Obsolete point to `SetBoxAuthAuthorizeApiUri` method instead of a field?

antusus (commented):
I'm not yet sure why we want to separate authorisation and revoke URL from base URL. If a customer has his own domain or subdomain within box would they change just base urls for API calls and leave the authorisation URLs not changed?

antusus (dismissed):
Great work!

antusus (approved):
👍

Issue #710, opened 2020-11-17
SDK stops working throwing System.TimeoutException for all requests after random time


ChristianGalla:
- [x] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [x] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [x] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
   - I think this is the same issue: [APIs-fail-with-TaskCanceledException](https://support.box.com/hc/en-us/community/posts/360049152914-APIs-fail-with-TaskCanceledException)
- [x] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue

We are using the SDK in a long running console application to upload many files to Box. Many copies of the application are running at the same time.

After some random time, it is not possible to download files anymore and all calls of BoxClient.FilesManager.DownloadAsync are throwing System.TimeoutException.

Other application instances are not directly affected but may fail at another time.

### Steps to Reproduce

Randomly after some hours, days or even weeks the API stops working and all calls of BoxClient.FilesManager.DownloadAsync are throwing System.TimeoutException.

This happens even for small csv files. It is possible to access the files via Box website even between two calls of this function.

Without success I have tried to retry calls to this method after some seconds delay. Even manual recreation of the Box session did not help. Reconnecting to Box is not throwing an error.

The only workaround I have found is to restart the application.

### Expected Behavior

The SDK keeps working.

### Error Message, Including Stack Trace

```
System.TimeoutException: Request timed out ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Box.V2.Request.HttpRequestHandler.<getResponse>d__7.MoveNext()
   --- End of inner exception stack trace ---
   at Box.V2.Request.HttpRequestHandler.<getResponse>d__7.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #819, opened 2022-04-29
fix: properly dispose response on exception to prevent connection pool starvation

mwwoda:
Fix for #710. When an exception is thrown for a stream response and the response is not properly disposed of, the connection is not properly closed and there is a possibility of starving the connection pool.


antusus (approved):
👍

PR #820, opened 2022-05-11
feat: add `version_number` to `BoxFileVersion`

### LINKED ISSUE #818: Provide a way to get the file version back from the PromoteVersionAsync method.
.Net 5.0 using nuget package Box.V2.Core 4.31.

When using FIlesManager.PromoteVersionAsync, the BoxFileVersion that is returned does not have a version property to know which version number the newly promoted version was assigned. There is a FileVersion property on the returned object, but it is null and there is no way to request it.

Either hydrate the FileVersion object property by default, or provide an additional property to the PromoteVersionAsync method to allow fields to include so you can specify to return that object.

I am currently forced to call the api again to get the versions and determine which version is the current one.

mwwoda:
Fixes #818

### DIFF
--- a/Box.V2.Test.Integration/Configuration/Commands/ICommand.cs
+++ b/Box.V2.Test.Integration/Configuration/Commands/ICommand.cs
@@ -43,4 +43,17 @@ namespace Box.V2.Test.Integration.Configuration.Commands
         /// <returns></returns>
         Task Execute(IBoxClient client);
     }
+
+    /// <summary>
+    /// Interface used to create and of resources. It does not perform dispose action. Usable in cases like new file version upload.
+    /// </summary>
+    public interface INonDisposableCommand : ICommand
+    {
+        /// <summary>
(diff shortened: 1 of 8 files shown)


Box.V2.Test.Integration/Configuration/Commands/ICommand.cs
```diff
@@ -43,4 +43,17 @@ public interface ICleanupCommand : ICommand
         /// <returns></returns>
         Task Execute(IBoxClient client);
     }
+
+    /// <summary>
+    /// Interface used to create and of resources. It does not perform dispose action. Usable in cases like new file version upload.
```
arjankowski, line 48: I think the `remove/dispose` phrase is missing after `and`

mwwoda: Godd catch!. I improved wording a bit.

arjankowski (approved):
👍

PR #821, opened 2022-05-12
chore: run tests under all supported major netcore versions

mwwoda:
I also refactored exponential backoff test to speed it up considerably.
Tests in core against different runtimes are in a single job, as this is faster than multiple ones.


arjankowski (approved):
👍

PR #830, opened 2022-05-25
docs: update docs after release

mwwoda:
Merge this when the next version is released (currently it is v4.3.1).

### DIFF
--- a/Box.V2.Core/Box.V2.Core.csproj
+++ b/Box.V2.Core/Box.V2.Core.csproj
@@ -3,14 +3,14 @@
   <PropertyGroup>
     <Description>Box.V2 for .Net Core</Description>
     <AssemblyTitle>Box.V2.Core</AssemblyTitle>
-    <Version>4.3.1</Version>
+    <Version>4.4.0</Version>
     <Authors>Box, Inc.</Authors>
 	<Owners>Box, Inc.</Owners>
 	<PackageProjectUrl>https://github.com/box/box-windows-sdk-v2</PackageProjectUrl>
   <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
 	<PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
 	<PackageDescription>Windows SDK for v2 of the Box API. The SDK is built for .NET Core and includes JWTAuth support.</PackageDescription>
-	<PackageReleaseNotes>See https://github.com/box/box-windows-sdk-v2/blob/main/CHANGELOG.md#431-2022-04-19</PackageReleaseNotes>
+	<PackageReleaseNotes>See https://github.com/box/box-windows-sdk-v2/blob/main/CHANGELOG.md#440-2022-06-14</PackageReleaseNotes>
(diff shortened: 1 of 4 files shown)


docs/files.md
```diff
@@ -260,18 +260,23 @@ A file can be unlocked by calling `FilesManager.UnLock(string id)` with the ID o
...
-Create a Shared Link
+Create or update a Shared Link
 --------------------
 
-A shared link for a file can be generated by calling
+A shared link for a file can be created or updated by calling
 `FilesManager.CreateSharedLinkAsync(string id, BoxSharedLinkRequest sharedLinkRequest, IEnumerable<string> fields = null)`.
 
-<!-- sample put_files_id_shared_link_create -->
+<!-- sample put_files_id shared_link_create -->
```
arjankowski, line 269: Are you sure that `put_files_id shared_link_create` is right here?

This page https://developer.box.com/reference/put-files-id--add-shared-link/ using this sample label 
`put_files_id add_shared_link`

And this page https://developer.box.com/reference/put-files-id--update-shared-link/ is using
`put_files_id update_shared_link`.

However when we want to add sample code here https://developer.box.com/guides/shared-links/create-or-update/, we should add `put_files_id add_shared_link` as well as in first example.

docs/folders.md
```diff
@@ -116,14 +116,14 @@ folder and all of its contents, set the optional `recursive` parameter to `true`
...
+Create or update a Shared Link
 ---------------------------------
 
-You can create a shared link for a folder by calling
+You can create or update a shared link for a folder by calling
 `FoldersManager.CreateSharedLinkAsync(string id, BoxSharedLinkRequest sharedLinkRequest, IEnumerable<string> fields = null)`
 with the ID of the folder and the shared link parameters.
 
-<!-- sample put_folders_id_shared_link_create -->
+<!-- sample put_folders_id shared_link_create -->
```
arjankowski, line 126: The same question as previously, is `put_folders_id shared_link_create` should be here?
Or maybe we should use `'put_folders_id add_shared_link` instead?

docs/web-links.md
```diff
@@ -79,16 +80,18 @@ Copy a Web Link
...
 
-Create a Shared Link
+Create or update a Shared Link
 --------------------
 
-A shared link for a web link can be generated by calling
+A shared link for a web link can be created or updated by calling
 `WebLinksManager.CreateSharedLinkAsync(string id, BoxSharedLinkRequest sharedLinkRequest, IEnumerable<string> fields = null)`.
 
+<!-- sample put_web_links_id add_shared_link -->
```
arjankowski, line 94: Here is OK

docs/files.md
mwwoda: You may be right. I took the labels from the original Jira ticket, but I see that perhaps we should use `put_files_id add_shared_link` instead (or both). I'll try to clarify this.

mwwoda: Fixed

docs/folders.md
mwwoda: Fixed

arjankowski (approved):
👍

PR #831, opened 2022-06-09
feat: add `can_edit` field to `SharedLink`

mwwoda:
I decided to model the SDK like open-api (can-edit for File, Folder and WebLink). An exception is thrown when you try to set `can_edit` `true` for Folder or WebLink. The API throws an exception when you do this, so we can throw it earlier.

Another possible solution would be to either separate SharedLink for File, Folder, and WebLink (a breaking change), or deprecate the existing methods in favor of new ones with new DTOs.


arjankowski (approved):
👍

PR #838, opened 2022-07-25
chore: replace semantic validation bot with action

mwwoda:
Using https://github.com/amannn/action-semantic-pull-request. Using `pull_request_target` trigger, so it will be run after the merge.


arjankowski (approved):
👍

PR #846, opened 2022-08-02
docs: Await folder creation

peterneave:
Async method must be awaited when creating a folder


mwwoda (approved):
Thanks for the contribution!

PR #850, opened 2022-08-23
feat: Add `content-type` field to sign request

congminh1254:
Closes: SDK-2252

### DIFF
--- a/Box.V2.Test/Fixtures/BoxSignRequest/GetSignRequest200.json
+++ b/Box.V2.Test/Fixtures/BoxSignRequest/GetSignRequest200.json
@@ -62,7 +62,8 @@
           "checkbox_value": true,
           "date_value": "2021-04-26T08:12:13.982Z",
           "type": "text",
-          "page_index": 4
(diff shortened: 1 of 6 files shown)


Box.V2.Test/Fixtures/BoxSignRequest/GetSignRequest200.json
```diff
@@ -62,7 +62,8 @@
           "checkbox_value": true,
           "date_value": "2021-04-26T08:12:13.982Z",
           "type": "text",
-          "page_index": 4
+          "page_index": 4,
+          "content_type": "checkbox"
```
mwwoda, line 66: This new field was added in the json, but we do not validate it in any way in the test. Unfortunately, for now we have to check each field separately. Could you add an assert checking this new field in at least one of the tests?

mwwoda (commented):
I left a comment regarding adding assert to tests.

mwwoda (approved):
LGTM!

PR #851, opened 2022-08-23
fix: fix null reference exception when it's not possible to get `runtime` version from the assembly

mwwoda:
If runtime version cannot be determined from assembly or in case of Blazor use the preferred way to get runtime version introduced in .net core 3.0

Also added check for AutomaticDecompression settings for HttpHandler, because it's not supported in Blazor at the moment.


antusus (approved):
👍

PR #856, opened 2022-09-08
feat: Add support for modifiable retention policies & enable deleting retention policy assignment

congminh1254:
Closes: SDK-2023

### DIFF
--- a/Box.V2/Models/BoxRetentionPolicy.cs
+++ b/Box.V2/Models/BoxRetentionPolicy.cs
@@ -20,6 +20,7 @@ namespace Box.V2.Models
         public const string FieldCanOwnerExtendRetention = "can_owner_extend_retention";
         public const string FieldAreOwnersNotified = "are_owners_notified";
         public const string FieldCustomNotificationRecipients = "custom_notification_recipients";
+        public const string FieldRetentionType = "retention_type";
 
         /// <summary>
         /// The name given to the retention policy
@@ -87,5 +88,11 @@ namespace Box.V2.Models
         /// </summary>
         [JsonProperty(PropertyName = FieldCustomNotificationRecipients)]
         public virtual List<BoxUser> CustomNotificationRecipients { get; set; }
+
+        /// <summary>
+        /// The type of retention policy. Value is one of modifiable or non-modifiable.
+        /// </summary>
+        [JsonProperty(PropertyName = FieldRetentionType)]
+        public virtual string RetentionType { get; set; }
     }
 }
--- a/Box.V2/Models/Request/BoxRetentionPolicyRequest.cs
+++ b/Box.V2/Models/Request/BoxRetentionPolicyRequest.cs
@@ -55,5 +55,12 @@ namespace Box.V2.Models.Request
         /// </summary>
         [JsonProperty(PropertyName = "custom_notification_recipients")]
         public List<BoxRequestEntity> CustomNotificationRecipients { get; set; }
+
+        /// <summary>
+        /// Used to determine the type of retention policy, value can be modifiable or non-modifiable
+        /// When updating a retention policy, you can use non-modifiable type only. You can convert a modifiable policy to non-modifiable, but not the other way around.
+        /// </summary>
+        [JsonProperty(PropertyName = "retention_type")]
+        public string RetentionType { get; set; }
     }
 }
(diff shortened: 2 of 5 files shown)


Box.V2.Test.Integration/Configuration/Commands/DisposableCommands/CreateRetentionPolicyCommand.cs
```diff
@@ -27,9 +28,27 @@ public async Task<string> Execute(IBoxClient client)
...
+                Policy = response;
+            }
+            catch
+            {
+                // TODO: 12-09-2022, @mcong
+                // There is an error on backend side, which will return 409 status code "conflict"
+                // but retention policy still created.
+                // Delete this try-catch after the issue is fixed.
+                var policies = await client.RetentionPoliciesManager.GetRetentionPoliciesAsync(_policyName);
+                if (policies.Entries.Count == 1)
```
mwwoda, line 43: That workaround will do for now. We should probably also do something when Count is not equal to 1 for some reason. Maybe throw an exception?

Box.V2/Models/BoxRetentionPolicy.cs
```diff
@@ -87,5 +88,11 @@ public class BoxRetentionPolicy : BoxEntity
         /// </summary>
         [JsonProperty(PropertyName = FieldCustomNotificationRecipients)]
         public virtual List<BoxUser> CustomNotificationRecipients { get; set; }
+
+        /// <summary>
+        /// The type of retention policy. Value is one of modifiable or non-modifiable.
+        /// </summary>
+        [JsonProperty(PropertyName = FieldRetentionType)]
+        public virtual string RetentionType { get; set; }
```
mwwoda, line 96: I think that we can use enum there if we have only 2 possible values.

Box.V2/Models/Request/BoxRetentionPolicyRequest.cs
```diff
@@ -55,5 +55,12 @@ public class BoxRetentionPolicyRequest
         /// </summary>
         [JsonProperty(PropertyName = "custom_notification_recipients")]
         public List<BoxRequestEntity> CustomNotificationRecipients { get; set; }
+
+        /// <summary>
+        /// Used to determine the type of retention policy, value can be modifiable or non-modifiable
+        /// When updating a retention policy, you can use non-modifiable type only. You can convert a modifiable policy to non-modifiable, but not the other way around.
+        /// </summary>
+        [JsonProperty(PropertyName = "retention_type")]
+        public string RetentionType { get; set; }
```
mwwoda, line 64: I think that we can use enum there if we have only 2 possible values.

mwwoda (commented):
I left a few comments

Box.V2.Test.Integration/BoxRetentionPolicyManagerIntegrationTest.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Box.V2.Test.Integration
+{
+    [TestClass]
+    public class BoxRetentionPolicyManagerIntegrationTest : TestInFolder
+    {
+        [TestMethod]
+        public async Task CreateRetentionPolicyAssignmentAsync_ForRetentionPolicyAssignmentRequest_ShouldSuccess()
+        {
+            var adminFolder = await CreateFolderAsAdmin("0");
```
mwwoda, line 17: Why do we need 2 folders there?

Box.V2.Test.Integration/BoxRetentionPolicyManagerIntegrationTest.cs
```diff
@@ -0,0 +1,37 @@
...
+            var adminFolder2 = await CreateFolderAsAdmin("0");
```
mwwoda, line 18: Why do we need 2 folders there?

Box.V2/Models/Request/BoxRetentionPolicyRequest.cs
congminh1254: I can only add it into `Box.V2/Config/Constants.cs`, it can not be added into enum due to the name `modifiable` and `non-modifiable`.

Box.V2.Test.Integration/Configuration/Commands/DisposableCommands/CreateRetentionPolicyCommand.cs
congminh1254: Fixed. I throw the same exception if the policy count is other than 1.

Box.V2/Models/Request/BoxRetentionPolicyRequest.cs
mwwoda: I believe that you still should be able to express this as an enum using `[EnumMember(Value = "non-modifiable)] `attribute. See `BoxSignRequestStatus` for example

mwwoda (approved):
LGTM!

PR #875, opened 2022-12-14
fix!: return proper object type from `GetFileVersionsUnderRetentionForAssignmentAsync` method

mwwoda:


antusus (approved):
I would only consider adding some IT tests for getting files under retention.

mwwoda:
> I would only consider adding some IT tests for getting files under retention.

Setting up the account to work with this endpoint was a hassle, so we will skip the integration test for this endpoint for now.

PR #876, opened 2022-12-14
chore: ignore `bug fixes` in `readme` during linting

mwwoda:
It seems that linting for multiple words cannot be excluded in the terminology rules we use https://github.com/sapegin/textlint-rule-terminology/issues/21. We can simply ignore the line that causes the problem with a comment, so we don't need to change the readme, since it is similar across SDKs.


antusus (approved):
👍

PR #883, opened 2023-01-12
feat: `BoxCCGAuth` add User and Admin clients factory methods without initial token

### LINKED ISSUE #879: Server Auth with CCG - initial setup for DI container registration
- [x] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [x] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [x] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [x] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue
The code provided in the [documentation example](https://github.com/box/box-windows-sdk-v2/blob/main/docs/authentication.md#server-auth-with-ccg) is working fine but most modern applications use DI with container registrations that are synchronous. [This PR](https://github.com/box/box-windows-sdk-v2/pull/799) added support for the CCG auth but requires a user token that is available via API call only and this is the root of the issue unless you do the API call to obtain a user token in a synchronous way, there is no way of registering e.g. `IBoxClient` into a container, so you have to wrap it into an async factory method.

Another gotcha is the token refresh via `event`, even if would wrap `IBoxClient` into an async factory method, register a method to the `IBoxClient.Auth.SessionAuthenticated` event that handles the token refresh (token caching), do some API calls (logic), I *should* unregister the method to avoid leaks.

### Steps to Reproduce
```csharp
var boxConfig = new BoxConfigBuilder("YOUR_CLIENT_ID", "YOUR_CLIENT_SECRET")
                .Build();
var boxCCG = new BoxCCGAuth(boxConfig);
var userToken = await boxCCG.UserTokenAsync("USER_ID"); //valid for 60 minutes so should be cached and re-used
IBoxClient userClient = boxCCG.UserClient(userToken, "USER_ID");
userClient.Auth.SessionAuthenticated += delegate(object o, SessionAuthenticatedEventArgs e)
{
    string newAccessToken = e.Session.AccessToken;
    // cache the new access token
};
```

refactored code with token caching and exposing `IBoxClient` via the async factory method
``` csharp
string? userToken = null;
async ValueTask<IBoxClient> CreateBoxClientAsync()
{
    userToken ??= await boxCCG.UserTokenAsync(userId); //valid for 60 minutes so should be cached and re-used
    var userClient = boxCCG.UserClient(userToken, serviceAccount);
    userClient.Auth.SessionAuthenticated += OnSessionAuthenticated;
    return userClient;
}

void OnSessionAuthenticated(object sender, SessionAuthenticatedEventArgs e)
{
    userToken = e.Session.AccessToken;
}

var userClient = await CreateBoxClientAsync();
// ... do logic ...
// I should still unregister the method, which is too much in my view
userClient.Auth.SessionAuthenticated -= OnSessionAuthenticated;
```

### Expected Behavior
Ability to instantiate `IBoxClient` without initial `user id => user token` exchange.

### Versions Used
.Net SDK: 4.6.0

[sdk-docs]: ./docs
[dev-forums]: https://community.box.com/t5/Platform-and-Development-Forum/bd-p/DeveloperForum
[github-repo]: https://github.com/box/box-windows-sdk-v2/search?type=Issues

JanVargovsky:
Fixes #879

### DIFF
--- a/Box.V2/Managers/BoxResourceManager.cs
+++ b/Box.V2/Managers/BoxResourceManager.cs
@@ -73,7 +73,7 @@ namespace Box.V2.Managers
             where T : class
         {
             AddDefaultHeaders(request);
-            await AddAuthorizationAsync(request);
+            await AddAuthorizationAsync(request).ConfigureAwait(false);
             var response = await ExecuteRequest<T>(request, queueRequest).ConfigureAwait(false);
 
             return converter != null ? response.ParseResults(converter) : response.ParseResults(_converter);
--- a/Box.V2/Managers/BoxResourceManager.cs
+++ b/Box.V2/Managers/BoxResourceManager.cs
@@ -14,7 +14,6 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Reflection;
-using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
@@ -74,7 +73,7 @@ namespace Box.V2.Managers
             where T : class
         {
             AddDefaultHeaders(request);
-            AddAuthorization(request);
+            await AddAuthorizationAsync(request);
             var response = await ExecuteRequest<T>(request, queueRequest).ConfigureAwait(false);
 
             return converter != null ? response.ParseResults(converter) : response.ParseResults(_converter);
@@ -108,13 +107,14 @@ namespace Box.V2.Managers
         {
             OAuthSession newSession = await _auth.RefreshAccessTokenAsync(request.Authorization).ConfigureAwait(false);
             AddDefaultHeaders(request);
-            AddAuthorization(request, newSession.AccessToken);
+            await AddAuthorizationAsync(request, newSession.AccessToken);
             return await _service.ToResponseAsync<T>(request).ConfigureAwait(false);
         }
 
-        protected void AddAuthorization(IBoxRequest request, string accessToken = null)
+        protected async Task AddAuthorizationAsync(IBoxRequest request, string accessToken = null)
         {
-            var auth = accessToken ?? _auth.Session.AccessToken;
+            var auth = accessToken ??
+                (_auth.Session ?? await _auth.RefreshAccessTokenAsync(null)).AccessToken;
 
             var authString = string.Format(CultureInfo.InvariantCulture, Constants.V2AuthString, auth);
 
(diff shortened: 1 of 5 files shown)


JanVargovsky:
I can update (simplify) the documentation if the code looks OK.

Box.V2/Managers/BoxResourceManager.cs
```diff
@@ -74,7 +73,7 @@ protected IBoxRequest AddDefaultHeaders(IBoxRequest request)
             where T : class
         {
             AddDefaultHeaders(request);
-            AddAuthorization(request);
+            await AddAuthorizationAsync(request);
```
mwwoda, line 76: please add `.ConfigureAwait(false)` here also

JanVargovsky:
I'm not the best at writing docu, so hopefully, it makes sense. There were invalid links, so I've updated them. You can review it now.

Do I have to fix something myself to make integration tests pass? based on CI log it seems like it's not configured.

docs/authentication.md
```diff
@@ -130,6 +130,10 @@ var boxConfig = new BoxConfigBuilder("YOUR_CLIENT_ID", "YOUR_CLIENT_SECRET")
                 .SetEnterpriseId("YOUR_ENTERPRISE_ID")
                 .Build();
 var boxCCG = new BoxCCGAuth(boxConfig);
+```
+
+There are two ways to create an admin client, first one used explicit admin token:
```
mwwoda, line 134: Small typos. I think you can change it to:
```There are two ways to create an admin client, the first one uses explicit admin token:```

mwwoda (commented):
Nice catch with old links! Just a small typo and we are ready to go 🚀

mwwoda:
> Do I have to fix something myself to make integration tests pass? based on CI log it seems like it's not configured.

The integration tests use some preconfigured account that calls the Box API directly. Since this account needs credentials, this workflow is restricted to internal collaborators to prevent malicious code that could be triggered by the fork. Tests will be run on the main branch when we merge this PR.

mwwoda:
Thanks @JanVargovsky once again for your input. It's always nice to see contributions from the community. I'll try to release a new package to nuget with these changes early next week to make it easier to access this feature. Have a great weekend!

JanVargovsky:
I get the security perspective, I just thought that it would pass, you just have to run it manually. Thanks for the review :)

PR #888, opened 2023-02-15
feat: add `start_date_field` and `description` to retention policies

mwwoda:


Box.V2/Models/BoxRetentionPolicyAssignment.cs
```diff
@@ -44,5 +45,11 @@ public class BoxRetentionPolicyAssignment : BoxEntity
         /// </summary>
         [JsonProperty(PropertyName = FieldFilterFields)]
         public virtual List<BoxMetadataFieldFilter> FilterFields { get; set; }
+
+        /// <summary>
+        /// The date the retention policy assignment begins.
```
arjankowski, line 50: I found two different descriptions for this field depending on the document I looked at:

`Attribute key to look for in the source for the retention start date` 

or

`The Metadata field GUID which will be used to specify the start date for the retention policy.`

What do you think about using one of them to to accurately describe this field?

Box.V2/Models/Request/BoxRetentionPolicyAssignmentRequest.cs
```diff
@@ -22,5 +22,11 @@ public class BoxRetentionPolicyAssignmentRequest
         /// </summary>
         [JsonProperty(PropertyName = "filter_fields")]
         public List<object> FilterFields { get; set; }
+
+        /// <summary>
+        /// The date the retention policy assignment begins.
```
arjankowski, line 27: Same here

mwwoda: ditto

arjankowski (approved):
👍

PR #909, opened 2023-05-26
fix: replace deprecated `BouncyCastle`

TFoxik:
Replace deprecated BouncyCastle.Crypto with BouncyCastle.Cryptography


mwwoda:
Thanks for your contribution. So fast! Looks like we can go with this PR safely. I'll just check a few things and merge it. 
We will consider this as a fix to bump the patch version of the package and include the change in the change log.

mwwoda (approved):

mwwoda:
Hi @jmallingergpr
5.3.0 with this change was published a moment ago. It should appear in nuget soon.

PR #923, opened 2023-09-19
feat: Add shared link support in file and folder managers

arjankowski:


bartlomiejleszczynski:
Minor thing, but: branch is referring to SDK-3342, but the issue is SDK-3341. The correct issue is the latter one in case some one is searching for it ;-)

mwwoda (approved):
🚀

PR #930, opened 2023-11-29
fix: support single object value in `BoxConflictErrorContextInfo`

mwwoda:
In case of `copyFolder` conflict exception single value is returned in `conflicts` parts of an exception. Added a serializer that deserializes from both arrays and objects into an existing collection of `conflicts`.


congminh1254 (approved):
LGTM :+1:

PR #945, opened 2024-04-10
fix: do not recreate Random each time

### LINKED ISSUE #944: ExponentialBackoff strategy recreates Random instance each time
- [X] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [X] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [X] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [X] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue
Within the ExponentialBackoff retry strategy, the `Random` instance is recreated each time `GetRetryTimeout` is called. This is an anti-pattern, and incorrect for .NET Framework based projects:
> In .NET Framework, the default seed value is derived from the system clock, which has finite resolution. As a result, different [Random](https://learn.microsoft.com/en-us/dotnet/api/system.random?view=net-8.0) objects that are created in close succession by a call to the parameterless constructor have identical default seed values and, therefore, produce identical sets of random numbers. You can avoid this problem by using a single [Random](https://learn.microsoft.com/en-us/dotnet/api/system.random?view=net-8.0) object to generate all random numbers. You can also work around it by generating your own random seed value and passing it to the [Random(Int32)](https://learn.microsoft.com/en-us/dotnet/api/system.random.-ctor?view=net-8.0#system-random-ctor(system-int32)) constructor. For more information, see the [Random(Int32)](https://learn.microsoft.com/en-us/dotnet/api/system.random.-ctor?view=net-8.0#system-random-ctor(system-int32)) constructor.

https://learn.microsoft.com/en-us/dotnet/api/system.random.-ctor

Note: this does not apply to Box.V2.Core, but they share the same repository.

### Steps to Reproduce
1. Execute multiple `ExponentialBackoff.GetRetryTimeout` calls in a row on .NET Framework 4.6.2+ targets

### Expected Behavior
A single `Random` instance is used in accordance with .NET Framework best practices.

N/A

### Screenshots
N/A

### Versions Used
.Net SDK: Box.V2 5.7.0
Windows: N/A

[sdk-docs]: ./docs
[dev-forums]: https://community.box.com/t5/Platform-and-Development-Forum/bd-p/DeveloperForum
[github-repo]: https://github.com/box/box-windows-sdk-v2/search?type=Issues

watfordsuzy:
On .NET Framework it is incorrect to recreate Random instances each time, these should instead be one per instance. On .NET Core and later this is not a limitation, but the libraries share a codebase.

Fixes #944

### DIFF
--- a/Box.V2/Box.V2.csproj
+++ b/Box.V2/Box.V2.csproj
@@ -276,6 +276,7 @@
     <Compile Include="Utility\ContentTypeMapper.cs" />
     <Compile Include="Utility\Retry.cs" />
     <Compile Include="Utility\SharedLinkUtils.cs" />
+    <Compile Include="Utility\ThreadSafeRandom.cs" />
     <Compile Include="Wrappers\BoxError.cs" />
     <Compile Include="Wrappers\BoxErrorContextInfo.cs" />
     <Compile Include="Wrappers\BoxBinaryRequest.cs" />
--- a/Box.V2/Utility/ExponentialBackoff.cs
+++ b/Box.V2/Utility/ExponentialBackoff.cs
@@ -4,8 +4,6 @@ namespace Box.V2.Utility
 {
     public class ExponentialBackoff : IRetryStrategy
     {
-        private readonly Random _random = new Random();
-
         public TimeSpan GetRetryTimeout(int numRetries)
         {
             var baseInterval = TimeSpan.FromSeconds(2.0);
@@ -13,7 +11,9 @@ namespace Box.V2.Utility
             var minRandomization = 1 - RETRY_RANDOMIZATION_FACTOR;
             var maxRandomization = 1 + RETRY_RANDOMIZATION_FACTOR;
 
-            var randomization = _random.NextDouble() * (maxRandomization - minRandomization) + minRandomization;
+            var randomization = ThreadSafeRandom.Instance.NextDouble()
+                * (maxRandomization - minRandomization) + minRandomization;
+
             var exponential = Math.Pow(2, numRetries - 1);
             var result = Math.Ceiling(exponential * baseInterval.TotalSeconds * randomization);
             return TimeSpan.FromSeconds(result);
--- a/Box.V2/Utility/ExponentialBackoff.cs
+++ b/Box.V2/Utility/ExponentialBackoff.cs
@@ -4,15 +4,16 @@ namespace Box.V2.Utility
 {
     public class ExponentialBackoff : IRetryStrategy
     {
+        private readonly Random _random = new Random();
+
         public TimeSpan GetRetryTimeout(int numRetries)
         {
             var baseInterval = TimeSpan.FromSeconds(2.0);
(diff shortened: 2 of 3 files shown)


mwwoda:
Hi @watfordsuzy 
Thanks for the contribution! I think it's a good practice in general reuse the `Random` instance so we could do it for both of the SDKs (Framework and Core). I think thread safety could be a potential issue there (e.g. in chunked file uploads) https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-random#thread-safety. What do you think?

watfordsuzy:
For .NET Core and beyond you don't strictly need to re-use it (but should) as the underlying implementation already gives you a Thread Safe implementation:
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Random.cs

On .NET Framework a thread local Random could be used instead, but there are some interesting caveats if you go and read the comments regarding `ThreadSafeRandom` used in .NET Core+.

mwwoda:
There is also an explicit thread safe implementation for `Random` https://learn.microsoft.com/en-us/dotnet/api/system.random.shared?view=net-8.0, but only for .NET 6+ and we need to support older versions too.

If you want to work on this further, I think a thread-safe solution is the way to go. The overhead of potential locks shouldn't be a big deal, since this feature is already used to delay subsequent api calls, so that would just mean a slight bigger delay. While arguably making it not thread-safe isn't such a big deal either, it might result in slightly different behavior in multi-threaded environment compared to the single-threaded one. If necessary it's also possible to provide two implementation for both SDKs with pragmas. See example https://github.com/box/box-windows-sdk-v2/blob/b410dfd7177528d02f243596ad0cae59c5a8c1a7/Box.V2/Request/HttpRequestHandler.cs#L286

watfordsuzy:
I've gone ahead and made the suggested change, adapting their internal `ThreadSafeRandom` to fit the existing code. I've also added a section for .NET 6+, delegating to the inboxed thread safe random implementation.

watfordsuzy:
I don't believe the Integration Test failures are related @mwwoda

mwwoda:
Don't worry about the integration tests. Unfortunately, we cannot currently run them on forks.

mwwoda:
The code looks pretty good. I see that you have introduced a modern implementation for .NET6+ too which is great. Unfortunately, I just checked that when you use a conditional compilation flag such as `NET6_0_OR_GREATER` when not explicitly targeting this version in the library, this code will be never used during compilation (so it is basically dead code at this point). We would have to explicitly target .NET 6 version, which I don't think we will tackle in the near future. Maybe you could remove implementation for .NET 6+, and then we can merge your change. Sorry for the confusion.

watfordsuzy:
@mwwoda no worries!

PR #972, opened 2025-05-27
chore: convert projects to modern sdk-style projects

mwwoda:


Box.V2/Box.V2.csproj
```diff
@@ -1,383 +1,45 @@
...
+    <TargetFramework>net462</TargetFramework>
+    <Version>5.8.0</Version>
+    <Description>Box SDK for v2 of the Box API. The SDK is targeting .NET Framework 4.6.2</Description>
+    <Authors>Box Inc.</Authors>
+    <RepositoryUrl>https://github.com/box/box-windows-sdk-v2.git</RepositoryUrl>
+    <RepositoryType>git</RepositoryType>
+    <PackageProjectUrl>https://github.com/box/box-windows-sdk-v2</PackageProjectUrl>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <PackageReleaseNotes>See https://github.com/box/box-windows-sdk-v2/blob/main/CHANGELOG.md#580-2024-07-22</PackageReleaseNotes>
+    <PackageTags>Box V2 SDK Platform Enterprise Collaboration Storage File Management</PackageTags>
```
arjankowski, line 13: In `Box.V2.Core/Box.V2.Core.csproj` file in `PackageTags` attribute the tags are separated by semicolon `;`.
Does it matter?

Shouldn't we add here `AssembyTitle` like in `Box.V2.Core.cproj`?

Box.V2.Core/Box.V2.Core.csproj
```diff
@@ -1,24 +1,18 @@
...
-	<PackageReleaseNotes>See https://github.com/box/box-windows-sdk-v2/blob/main/CHANGELOG.md#580-2024-07-22</PackageReleaseNotes>
-	<Copyright>Copyright 2021</Copyright>
-	<PackageTags>Box;V2;SDK;Platform;Enterprise;Collaboration;Storage;File Management</PackageTags>
+    <PackageId>Box.V2.Core</PackageId>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <Version>5.8.0</Version>
+    <Description>Box SDK for v2 of the Box API. The SDK is targeting .Net Standard 2.0</Description>
+    <Authors>Box Inc.</Authors>
+    <AssemblyTitle>Box.V2.Core</AssemblyTitle>
+	  <PackageProjectUrl>https://github.com/box/box-windows-sdk-v2</PackageProjectUrl>
```
arjankowski, line 10: In Box.V2.csproj there is also a `RepositoryUrl` tag `<RepositoryUrl>https://github.com/box/box-windows-sdk-v2.git</RepositoryUrl>`. Shouldn't we add it here as well?

Box.V2/Box.V2.csproj
mwwoda: Looks like we can safely remove:
- `AssemblyName` - 
> Specify the full name and extension of the file you want to create. If you don't specify the name of the output file, MSBuild uses the name of the project to specify the name of the output assembly. 
Taken from https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/output#outputassembly

- `AssemblyTitle` - it's inferred from the `AssemblyName`.

After removing these props, looks like both title and name have proper values 

Screenshot 2025-05-29 at 11 49 15

Box.V2.Core/Box.V2.Core.csproj
mwwoda: Good catch! I'll add it

mwwoda: Hmm, looks like it's a SourceLink property and it is used to set `RepositoryUrl` anyway https://github.com/dotnet/sourcelink/issues/659#issuecomment-711288537. So I don't think we need to add it here.

PR #974, opened 2025-07-21
feat: Allow injecting private key decryptor for jwt

mwwoda:


Box.V2/JWTAuth/DefaultPrivateKeyDecryptor.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                throw new BoxCodingException("Invalid private key!");
+            }
+
+            RSA rsa = null;
+            if (privateKey is AsymmetricCipherKeyPair)
+            {
+                var ackp = (AsymmetricCipherKeyPair)privateKey;
+
+                rsa = ToRSA((RsaPrivateCrtKeyParameters)ackp.Private);
```
Pask423, line 42: how about removing the rsa variable and returning the result directly

Pask423 (approved):

